PR #2, opened 2018-08-09
Project Scaffolding

virkt25:
This creates the scaffolding for this example. The app was generated using `lb4 shopping`. Subsequent changes to the application are documented based on commits (should also make review easier). 

**COMMITS**
- feat: project scaffolding using lb4 app
- chore: add license, license headers for code files
- style: add markdown and commit message linting
- chore: update .gitignore
- docs: update README.md
- ci: add appveyor and travis config


package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "shopping",
```
raymondfeng: `@loopback/example-shopping`?

raymondfeng:
I'm curious why this example is not added to `loopback-next/examples`.

virkt25:
Well, this repository will be connected to a database (MongoDB most likely) for testing on Travis which could significantly slow down the builds for `loopback-next`. 

@bajtos comment from https://github.com/strongloop/loopback-next/issues/1476#issuecomment-401371422

> With that said, there is still a question whether this example should live in our main loopback-next monorepo, or in its own GitHub repo. Considering the cost of setting up a Redis instance, a database server, etc. I think we should keep this example out of our main monorepo. I don't want to start a Docker container with Redis every time I test a change in loopback-next!

raymondfeng:
That's fair.

src/application.ts
```diff
@@ -0,0 +1,42 @@
...
+    };
+  }
+
+  async start() {
+    await super.start();
+
+    const server = await this.getServer(RestServer);
+    const port = await server.get(RestBindings.PORT);
+    console.log(`Server is running at http://127.0.0.1:${port}`);
+    console.log(`Try http://127.0.0.1:${port}/ping`);
```
bajtos: IIRC, this template has been updated recently to use `this.restServer.url`, see https://github.com/strongloop/loopback-next/pull/1572. Could you please update the implementation of `start` method accordingly? Feel free to do so in a follow-up pull request, I think it will be easier to review.

PR #4, opened 2018-08-16
feat: adding artifacts for shopping cart

raymondfeng:
Add model/datasource/repository for redis backed shopping carts.

See https://github.com/strongloop/loopback-next/issues/1481


test/integration/shopping-cart.repository.integration.ts
```diff
@@ -0,0 +1,101 @@
...
+        PortBindings: {
+          '6379/tcp': [
+            {
+              HostPort: '6379',
+            },
+          ],
+        },
+      },
+    });
+    await container.start();
```
bajtos: How long does it take to start a Redis container?

I am concerned about the overhead this may be adding to `npm t`. When developing locally, we don't want to (re)start the Redis instance every time we want to re-run our test suite. 

Instead, we should start any database instances beforehand. When running on CI, the CI build script should start the instances before running the tests - Travis CI provides hooks for that.

See also https://github.com/strongloop/loopback4-example-shopping/pull/5. Since Travis CI already supports Redis, I think we should use the same approach from #5 and use Travis-CI provided instance of Redist instead of using Docker.

Related best practice: [Clean the database before each test](https://loopback.io/doc/en/lb4/Testing-your-application.html#clean-the-database-before-each-test).

test/integration/shopping-cart.repository.integration.ts
```diff
@@ -0,0 +1,101 @@
...
+    await container.stop();
+    await container.remove();
+  });
+
+  before(() => {
+    cart1 = new ShoppingCart({
+      userId: 'u01',
+      items: [
+        {
+          productId: 'p1',
```
bajtos: Please follow our best practices, in particular [Use data builders](https://loopback.io/doc/en/lb4/Testing-your-application.html#use-test-data-builders).

raymondfeng:
@bajtos @virkt25 PTAL

src/repositories/shopping-cart.repository.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import {DefaultKeyValueRepository} from '@loopback/repository';
+import {ShoppingCart} from '../models/shopping-cart.model';
+import {RedisDataSource} from '../datasources/redis.datasource';
+
+export class ShoppingCartRepository extends DefaultKeyValueRepository<
+  ShoppingCart
+> {
+  constructor() {
+    super(ShoppingCart, new RedisDataSource());
```
virkt25: The DataSource should be injected.

bin/start-dbs.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+if [ -z "$CI" ]; then
+  MONGO_CONTAINER_NAME="mongodb_c"
+  docker rm -f $MONGO_CONTAINER_NAME
```
jannyHou: is it a dup to remove it again in the start file? I notice that the same cmd is called in `stop-dbs.sh`.

bin/stop-dbs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
jannyHou: should travis call this file after test?

src/models/shopping-cart.model.ts
```diff
@@ -0,0 +1,50 @@
...
+// Node module: @loopback/example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+/**
+ * Item in a shopping cart
+ */
+export class ShoppingCartItem {
```
jannyHou: I am afraid we don't generate OpenAPI schema for a class without decorating it with `@model` and `@property`.

bin/stop-dbs.sh
raymondfeng: Travis recycles the containers. We don't have to worry about the `stop`. BTW, our Travis Mac build uses `brew`.

bin/start-dbs.sh
raymondfeng: It's just to be sure that such containers are removed. We can call `./stop-dbs.sh` instead.

package.json
```diff
@@ -20,7 +20,7 @@
     "clean": "lb-clean dist*",
     "commit": "git-cz",
     "commitmsg": "commitlint -E GIT_PARAMS",
-    "docker": "./bin/setup.sh",
+    "docker": "./bin/start-dbs.sh",
```
virkt25: Can we add a `docker:stop` which calls `stop-dbs.sh`.

bin/stop-dbs.sh
virkt25: No, Travis pretty much kills the instance that runs the tests once the tests are done ... and we aren't starting a docker containre on Travis but a service.

bin/start-dbs.sh
jannyHou: > Not really, I added stop-dbs.sh so that I can shut down these containers manually for local tests.

Oh I see 👌

PR #5, opened 2018-08-16
feat: implement basic user component

virkt25:
implements the remainder of https://github.com/strongloop/loopback-next/issues/1480


package.json
```diff
@@ -50,13 +50,17 @@
...
     "@loopback/boot": "^0.12.2",
+    "@loopback/context": "^0.12.2",
     "@loopback/core": "^0.11.2",
     "@loopback/dist-util": "^0.3.5",
+    "@loopback/openapi-v3": "^0.12.2",
     "@loopback/repository": "^0.14.2",
     "@loopback/rest": "^0.19.2",
-    "@loopback/openapi-v3": "^0.12.2"
+    "@types/bcrypt": "^2.0.0",
+    "bcrypt": "^3.0.0",
```
raymondfeng: Please use https://github.com/dcodeIO/bcrypt.js instead.

raymondfeng: `bcrypt` is a binary add-on and it requires compilation.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -0,0 +1,127 @@
...
+    await app.stop();
+  });
+
+  it('creates new user when POST /users is invoked', async () => {
+    await client
+      .post('/users')
+      .send(user)
+      .expect(200)
+      // tslint:disable-next-line:no-any
+      .expect(function(res: any) {
```
shimks: it's not all that important, but we can use the supertest's promise support here:
```ts
const response = await client.post('/users').send(user).expect(200);
// assertions
```

src/datasources/db.datasource.ts
```diff
@@ -0,0 +1,14 @@
+import {inject} from '@loopback/core';
```
shimks: missing license header here

package.json
virkt25: And we don't want that? CI was failing before because I was running install with `--ignore-scripts`. The JS version is about 30% slower according to their own docs.

src/controllers/user.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  @post('/users')
+  async create(@requestBody() obj: User): Promise<User> {
+    // Validate Email
+    if (!isemail.validate(obj.email)) {
+      throw new HttpErrors.UnprocessableEntity('invalid email');
+    }
+
+    // Validate Password Length
+    if (obj.password.length < 8) {
```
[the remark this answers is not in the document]
virkt25: We don't have a default User model like we did in LB3 which is where these checks / hashing would've been implemented. So for now people would have the flexibility to implement `User` how they want.

package.json
raymondfeng: Plain JS module is much more desirable for an example app. It will remove a lot of support issues in the future.

src/controllers/user.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+const genSaltAsync = promisify(genSalt);
+const hashAsync = promisify(hash);
+
+export class UserController {
+  constructor(
+    @repository(UserRepository) public userRepository: UserRepository,
+  ) {}
+
+  @post('/users')
+  async create(@requestBody() obj: User): Promise<User> {
```
b-admike: nitpick: I suggest we rename `obj` to `userObj` or `user` or something along those lines

src/controllers/user.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+    // Save & Return Result
+    const user = await this.userRepository.create(obj);
+    delete user.password;
+    return user;
+  }
+
+  @get('/users/{id}')
+  async findById(@param.path.string('id') id: string): Promise<User> {
+    const user = await this.userRepository.findById(id);
+    delete user.password;
```
b-admike: I wonder if juggler can take care of this for us using some sort of flag.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  it('throws error for POST /users with a string', async () => {
+    await client
+      .post('/users')
+      .send('hello')
+      .expect(415);
+  });
+
+  it('returns a user with given id when POST /user/{id} is invoked', async () => {
```
shimks: POST -> GET

test/acceptance/user.controller.acceptance.ts
```diff
@@ -0,0 +1,120 @@
...
+    const userWithId = Object.assign({}, user, {id: userId});
```
shimks: Instead of setting `userId` in a separate test, I think it's better for the user to be created in the test (through the repository), set the `userId`, and then call the endpoint. This allows tests to not be dependent upon each other.

Also, the created user instances need to be deleted after each tests

src/controllers/user.controller.ts
virkt25: That's the default we generate using the `lb4 controller` command but a good call. Will update.

src/controllers/user.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+        'password must be minimum 8 characters',
+      );
+    }
+
+    // Salt + Hash Password
+    user.password = await hashAsync(user.password, 10);
+
+    // Save & Return Result
+    const savedUser = await this.userRepository.create(user);
+    delete savedUser.password;
```
bajtos: While this is good enough for the initial implementation, it is also very brittle. We should find a more robust way how to allow models to hide certain properties from `toJSON` output.

In the past, I had very good experience with moving the password to a different model (table/collection) and use `hasMany` relation. As a nice side effect, by keeping a hash of all previous passwords, we can easily implement a password policy like "cannot reuse the same password".

src/controllers/user.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+    return savedUser;
+  }
+
+  @get('/users/{id}')
+  async findById(@param.path.string('id') id: string): Promise<User> {
+    const user = await this.userRepository.findById(id, {
+      fields: {password: false},
```
bajtos: Ditto here.

test/acceptance/ping.controller.acceptance.ts
```diff
@@ -5,7 +5,7 @@
 
 import {createClientForHandler, supertest} from '@loopback/testlab';
 import {RestServer} from '@loopback/rest';
-import {ShoppingApplication} from '../';
+import {ShoppingApplication} from '../..';
```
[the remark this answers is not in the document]
bajtos: @shimks @virkt25 could you please create a follow up issue or open a pull request to fix the template? Otherwise we are most likely going to forget about this.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -0,0 +1,132 @@
...
+import {UserRepository} from '../../src/repositories';
+import {UserDataSource} from '../../src/datasources';
+
+describe('UserController', () => {
+  let app: ShoppingApplication;
+  let server: RestServer;
+  let client: supertest.SuperTest<supertest.Test>;
+  const userRepo = new UserRepository(new UserDataSource());
+
+  const user = {
```
bajtos: I'd prefer us to follow our own best practices, see [Use data builders](https://loopback.io/doc/en/lb4/Testing-your-application.html#use-test-data-builders).

Issue #8, opened 2018-09-12
high server vulnerability after npm install


matejthetree:
I cloned 93201921663b375120aed74456648166cf48d3d1

and I get this error in console after doing npm install

```
> opencollective postinstall


                       Thanks for installing commitizen 🙏
                 Please consider donating to our open collective
                        to help us maintain this package.

                            Number of contributors: 58
                               Number of backers: 3
                              Annual budget: US$ 34
                             Current balance: US$ 34

         👉  Donate: https://opencollective.com/commitizen/donate

added 933 packages from 1479 contributors and audited 5496 packages in 33.028s
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
matejs-MacBook-Pro:loopback4-example-shopping matejsimunic$ npm audit
npm ERR! code EAUDITNOLOCK
npm ERR! audit Neither npm-shrinkwrap.json nor package-lock.json found: Cannot audit a project without a lockfile
npm ERR! audit Try creating one first with: npm i --package-lock-only

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/matejsimunic/.npm/_logs/2018-09-12T06_05_25_918Z-debug.log
matejs-MacBook-Pro:loopback4-example-shopping matejsimunic$ npm i --package-lock-only
audited 5496 packages in 6.933s
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
matejs-MacBook-Pro:loopback4-example-shopping matejsimunic$ npm audit
npm ERR! code EAUDITNOLOCK
npm ERR! audit Neither npm-shrinkwrap.json nor package-lock.json found: Cannot audit a project without a lockfile
npm ERR! audit Try creating one first with: npm i --package-lock-only

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/matejsimunic/.npm/_logs/2018-09-12T06_05_53_385Z-debug.log

```

virkt25:
What version of Node / npm are you using? Can you try against `master` please and report back.

raymondfeng:
The example itself has `package-lock.json` disabled. As a result, `npm audit` won't work. You can remove `package-lock=false` from `.npmrc` and try again.

@virkt25 I'm wondering if `package-lock` should be enabled for applications.

raymondfeng:
See #9

virkt25:
@matejthetree Now that we have a package-lock.json file for this project, is this issue good to close?

PR #9, opened 2018-09-12
Upgrade deps

raymondfeng:
Fixes the `npm audit` complaint reported by https://github.com/strongloop/loopback4-example-shopping/issues/8

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-package-lock=false
+package-lock=true
--- a/package.json
+++ b/package.json
@@ -52,32 +52,32 @@
     "src"
   ],
   "dependencies": {
-    "@loopback/boot": "^0.12.2",
-    "@loopback/context": "^0.12.2",
-    "@loopback/core": "^0.11.2",
-    "@loopback/dist-util": "^0.3.5",
-    "@loopback/openapi-v3": "^0.12.2",
-    "@loopback/repository": "^0.15.0",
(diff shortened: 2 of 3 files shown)


.npmrc
```diff
@@ -1 +1 @@
-package-lock=false
+package-lock=true
```
virkt25: Can we just get rid of this file as the default behaviour is to create a lock file?

raymondfeng: Some developers (like me) have a global npm config to disable `package-lock`.

package.json
```diff
@@ -52,32 +52,32 @@
...
   "devDependencies": {
-    "@commitlint/cli": "^7.0.0",
-    "@commitlint/config-conventional": "^7.0.1",
-    "@commitlint/travis-cli": "^7.0.0",
-    "@loopback/build": "^0.7.0",
-    "@loopback/testlab": "^0.11.2",
+    "@commitlint/cli": "^7.1.2",
+    "@commitlint/config-conventional": "^7.1.2",
+    "@commitlint/travis-cli": "^7.1.2",
+    "@loopback/build": "^0.7.2",
```
virkt25: 0.7.3

package.json
```diff
@@ -52,32 +52,32 @@
...
+    "@loopback/testlab": "^0.12.1",
```
virkt25: 0.12.2

raymondfeng:
Close to favor #10

PR #11, opened 2018-09-14
feat: add shopping-cart controller

raymondfeng:


raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/1481

src/controllers/shopping-cart.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    @repository(ShoppingCartRepository)
+    public shoppingCartRepository: ShoppingCartRepository,
+  ) {}
+
+  @put('/shoppingCarts/{userId}')
+  async set(
+    @param.path.string('userId') userId: string,
+    @requestBody({description: 'shopping cart'}) cart: ShoppingCart,
+  ) {
+    console.log('Shopping cart received', cart);
```
bajtos: Forgotten `console.log` statement?

src/controllers/shopping-cart.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (userId !== cart.userId) {
+      throw new HttpErrors['400'](
```
bajtos: We usually use `HttpErrors.BadRequest` instead of `HttErrors['400']`.

src/models/shopping-cart.model.ts
```diff
@@ -8,18 +8,22 @@ import {Entity, model, property} from '@loopback/repository';
 /**
  * Item in a shopping cart
  */
+@model()
```
virkt25: We need to make this a model so this is registered with the openapi spec?

raymondfeng: Yes, that's what leads to https://github.com/strongloop/loopback-next/pull/1710 too.

PR #12, opened 2018-09-17
user order history component

virkt25:


src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+   */
+  @post('/users/{userId}/orders', {
+    responses: {
+      '200': {
+        description: 'User.Order model instance',
+        content: {'application/json': {'x-ts-type': Order}},
+      },
+    },
+  })
+  async create(
```
raymondfeng: For a controller, is `createOrder` a better name?

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  @patch('/users/{userId}/orders', {
+    responses: {
+      '200': {
+        description: 'User.Order PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async patch(
```
raymondfeng: I would like to see a name meaningful for business logic, such as `updateOrders`.

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Order}},
+          },
+        },
+      },
+    },
+  })
+  async find(
+    @param.path.string('userId') userId: string,
+    @param.query.string('filter') filter?: Filter,
```
raymondfeng: See https://github.com/strongloop/loopback-next/pull/1679

src/models/shopping-cart.model.ts
```diff
@@ -25,6 +25,10 @@ export class ShoppingCartItem {
    */
   @property()
   price?: number;
+
+  constructor(item: Partial<ShoppingCartItem>) {
+    Object.assign(this, item);
```
raymondfeng: Is `super(item);` better?

raymondfeng:
Travis is failing with the following error:
```
․․․․․Unhandled error in POST /users/5b9fb25c1185f50a892ab737/orders: 500 TypeError: Class constructor ShoppingCartItem cannot be invoked without 'new'
    at /home/travis/build/strongloop/loopback4-example-shopping/node_modules/loopback-datasource-juggler/lib/list.js:68:16
    at Array.forEach (<anonymous>)
    at new List (/home/travis/build/strongloop/loopback4-example-shopping/node_modules/loopback-datasource-juggler/lib/list.js:66:9)
    at List (/home/travis/build/strongloop/loopback4-example-shopping/node_modules/loopback-datasource-juggler/lib/list.js:17:12)
```
I think we need to fix juggler to support class as a model function.

virkt25:
> I think we need to fix juggler to support class as a model function.

See https://github.com/strongloop/loopback-datasource-juggler/pull/1624

raymondfeng:
The juggler issue is fixed. Please pick up loopback-datasource-juggler@3.24.0

raymondfeng:
Please sign off the commits.

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+  async create(
+    @param.path.string('userId') userId: string,
+    @requestBody() order: Order,
+  ): Promise<Order> {
+    if (userId !== order.userId) {
+      throw new HttpErrors.BadRequest(
+        `User id does not match: ${userId} !== ${order.userId}`,
+      );
+    }
+    delete order.userId;
```
bajtos: I find it suspicious that we have to delete `order.userId`. I would expect the hasMany relation repository to take care of that. What am I missing?

src/controllers/user-order.controller.ts
bajtos: `update` does not make it clear whether it's a full replace or a partial patch. Let's use `patchOrders` please.

src/datasources/order.datasource.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {inject} from '@loopback/core';
+import {juggler, AnyObject} from '@loopback/repository';
+const config = require('./mongo.datasource.json');
+
+export class OrderDataSource extends juggler.DataSource {
```
bajtos: Is there any particular reason for having different datasource classes for Customer and Order models? Typically, a single datasource should be shared by all models persisted by the same database.

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+        description: 'User.Order model instance',
+        content: {'application/json': {'x-ts-type': Order}},
+      },
+    },
+  })
+  async create(
+    @param.path.string('userId') userId: string,
+    @requestBody() order: Order,
+  ): Promise<Order> {
+    if (userId !== order.userId) {
```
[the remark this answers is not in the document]
virkt25: We don't need to as `requestBody` validation does that for us. If you make a request with an `order` not containing a `userId`, the following error is thrown: 

```
{
  "error": {
    "statusCode": 422,
    "name": "UnprocessableEntityError",
    "message": "The request body is invalid. See error object `details` property for more info.",
    "details": [
      {
        "path": "",
        "code": "required",
        "message": "should have required property 'userId'",
        "info": {
          "missingProperty": "userId"
        }
      }
    ]
  }
}
```

src/controllers/user-order.controller.ts
virkt25: Will update once it lands.

src/models/shopping-cart.model.ts
virkt25: This is not a derived class, there is no super.

src/controllers/user-order.controller.ts
virkt25: Related to https://github.com/strongloop/loopback-next/issues/1719. This is a shortcoming of the relation repository where it actually throws an error if the property is present. If I don't delete the `userId` myself, I get the following error: 

```
Unhandled error in POST /users/5ba12137d078ee5357c02b00/orders: 500 Error: Property "userId" cannot be changed!
    at Object.constrainDataObject (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/repository/src/repositories/constraint-utils.ts:59:13)
    at DefaultHasManyEntityCrudRepository.create (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/repository/src/repositories/relation.repository.ts:79:7)
    at UserOrderController.createOrders (/Users/taranveer/LoopBack/loopback4-example-shopping/src/controllers/user-order.controller.ts:48:47)
    at value_promise_1.transformValueOrPromise.args (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/context/src/resolver.ts:240:37)
    at Object.transformValueOrPromise (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/context/src/value-promise.ts:269:12)
    at Object.invokeMethod (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/context/src/resolver.ts:235:10)
    at ControllerRoute.invokeHandler (/Users/taranveer/LoopBack/loopback4-example-shopping/node_modules/@loopback/rest/src/router/routing-table.ts:418:18)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

src/datasources/order.datasource.ts
virkt25: Yes, the reason was to use the same connection information while setting a different database name in the Class for each model. 

That said, we default the database name based on the model so I've updated the PR to just have a single DataSource shared by all models.

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -58,7 +58,7 @@ describe('UserOrderController acceptance tests', () => {
 
     expect(res.body.orderId).to.be.a.String();
     delete res.body.orderId;
-    expect(res.body).to.deepEqual(order);
+    expect(res.body).to.deepEqual(JSON.parse(JSON.stringify(order)));
```
raymondfeng: `order.toJSON()` should do the trick, isn't it?

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -183,11 +183,11 @@ describe('UserOrderController acceptance tests', () => {
       userId: '',
       total: 99.99,
       products: [
-        {
+        new ShoppingCartItem({
```
raymondfeng: Why do we have to use `ShoppingCartItem` class?

src/datasources/order.datasource.ts
bajtos: > Yes, the reason was to use the same connection information while setting a different database name in the Class for each model.

I think you may be confusing database and table name?

In SQL world, a single database server can provide multiple different databases as a kind of a multi-tenancy. Each database can have multiple tables. In LoopBack, we map models to tables.

From my experience, a single application typically stores data in a single database. This make it easy to leverage SQL features like the JOIN statement. I am not sure is there are any SQL databases that would support JOIN statements across different databases.

test/acceptance/user-order.controller.acceptance.ts
bajtos: IIRC, `order.toJSON()` returns an object that includes properties with undefined values.

```
// our expectation
{
  title: 'My Todo',
  isCompleted: undefined
}
```

On the other hand, the actual JSON string send by the server and parsed by the client in the test leaves out properties set to `undefined`. 

```
// the actual value
{
  title: 'My Todo'
}
```

AFAIK, the deep equality check distinguishes between a property set to `undefined` and a property that does not exist on the object at all.

As a result, the expected object obtained via `.toJSON()` is  rejected as being different from the actual value parsed from a JSON string.

I have encountered this issue myself in https://github.com/strongloop/loopback-next/pull/1679, I am going to add a new `testlab` helper to deal with that. See https://github.com/strongloop/loopback-next/blob/b4bba3d09d663b2919bd6ca3c5a5430d70e1d8c7/packages/testlab/src/misc.ts

Based on this discussion, I am going to open a new pull request to add this helper only, so that we can start using it sooner.

bajtos: > Based on this discussion, I am going to open a new pull request to add this helper only, so that we can start using it sooner.

See https://github.com/strongloop/loopback-next/pull/1733

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+  before(setupApplication);
+  beforeEach(clearDatabase);
+  after(async () => {
+    await app.stop();
+  });
+
+  it('creates an order for a user with a given orderId', async () => {
+    const user = await givenAUser();
+    const userId = user.id.toString();
+    const order = {
```
bajtos: Have you considered using a test-data builder here and specify only few property values that are relevant for this test?

Acceptance tests are verifying that all application parts work together, they should not be verifying behavior of individual components (we have unit and integration tests for that). 

In this particular case, we should rely on integration tests to verify that all properties provided to `orderRepo.create` are correctly stored in the database and retrieved back. IMO, we should use `const order = givenAnOrder()` here, because the actual order details are not important for this test.

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+      .post(`/users/${userId}/orders`)
+      .send(order)
+      .expect(200, order);
+  });
+
+  it('creates an order for a user without a given orderId', async () => {
+    const user = await givenAUser();
+    const userId = user.id.toString();
+    const order = givenAOrder();
+    order.userId = userId;
```
bajtos: A test-data builder typically accepts a Partial object allowing callers to customize/overwrite default values provided by the builder.

```ts
const order = givenAnOrder({userId});
```

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+    order.userId = 'hello';
+
+    await client
+      .post(`/users/${userId}/orders`)
+      .set('Content-Type', 'application/json')
+      .send(order)
+      .expect(400);
+  });
+
+  it.skip('throws an error when creating an order for a non-existent user', async () => {});
```
bajtos: Please create a follow-up issue (if you haven't do so already) and add a code comment pointing to that issue. (Explain that this test is skipped because the linked issue is not implemented yet.)

Also note that the test-function is optional for both `it` and `it.skip`. This line can be simplified as follows:

```ts
it.skip('throws an error when creating an order for a non-existent user');
```

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+
+      const getOrder2 = await orderRepo.find({
+        where: {orderId: savedOrder2.orderId},
+      });
+      savedOrder2.total = 9.99;
+      savedOrder2.orderId = savedOrder2.orderId!.toString();
+      expect(getOrder2).to.deepEqual([savedOrder2]);
+    });
+
+    it.skip('patches orders matching filter for a given user', async () => {});
```
bajtos: Ditto.

Please create a follow-up issue (if you haven't do so already) and add a code comment pointing to that issue. (Explain that this test is skipped because the linked issue is not implemented yet.)

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+      });
+      expect(getOrder1).to.deepEqual([]);
+
+      const getOrder2 = await orderRepo.find({
+        where: {orderId: savedOrder2.orderId},
+      });
+      expect(getOrder2).to.deepEqual([]);
+    });
+
+    it.skip('deletes orders matching filter for a given user', () => {});
```
bajtos: This looks like a duplicated test case?

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    await app.boot();
+    await app.start();
+
+    client = createClientForHandler(server.requestHandler);
+  }
+
+  async function givenAUser() {
+    const user = {
+      email: 'test@loopback.io',
```
bajtos: Please use `@example.com` address. See https://loopback.io/doc/en/contrib/style-guide.html#email-examples

test/acceptance/user.controller.acceptance.ts
```diff
@@ -109,23 +97,27 @@ describe('UserController', () => {
...
+    delete newUser.orders;
     // MongoDB returns an id object we need to convert to string
     // since the REST API returns a string for the id property.
     newUser.id = newUser.id.toString();
 
     await client.get(`/users/${newUser.id}`).expect(200, newUser.toJSON());
   });
 
-  function givenAnApplication() {
+  async function setupApplication() {
```
bajtos: `setupApplication` is duplicated in acceptance tests. Can we extract a shared helper please?

I understand `setupApplication` is updating global variables. Can we leverage object destructuring as an easy-to-use solution?

```ts
// the helper
async function setupApplication() {
  // ...
  return {app, server, client};
}


// usage in tests
let app: Application;
let server: RestServer;
let client: supertest.SuperTest<supertest.Test>;

before(async () => {
  // notice no "const", "let" or "var"
  {app, server, client} = await setupApplication();
});
```

Also since the function is not setting only the app, but also the server and the client, I think we should use a different name, e.g. `setupAcceptanceWorkbench`.

test/acceptance/user-order.controller.acceptance.ts
bajtos: See also https://github.com/strongloop/loopback-next/pull/1734

test/acceptance/user-order.controller.acceptance.ts
virkt25: The first test is to test that all orders are deleted while this one would test deleting only 1 out of the 2 orders by passing an appropriate filter.

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+   */
+  @post('/users/{userId}/orders', {
+    responses: {
+      '200': {
+        description: 'User.Order model instance',
+        content: {'application/json': {'x-ts-type': Order}},
+      },
+    },
+  })
+  async createOrders(
```
raymondfeng: It should be `createOrder` as the method only accepts one order.

src/controllers/user-order.controller.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  @patch('/users/{userId}/orders', {
+    responses: {
+      '200': {
+        description: 'User.Order PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async patchOrders(
```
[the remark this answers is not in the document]
virkt25: Yea, this builds on the controller we use for `todo-list` relational controller. We should probably introduce singular controllers for these operations where the orderId is known. 

`/users/{userId}/orders/{orderId}` -- Will make an issue.

bajtos:
>We should probably introduce singular controllers for these operations where the orderId is known. 
> 
>`/users/{userId}/orders/{orderId}` 

We have discussed this in the past, when implementing "hasMany" relation. Updating a single model via relation endpoint does not make much sense. Once you know the id of the target model (`orderId`), you can update the model directly via non-relation API (`PATCH /orders/{orderId}`).

The endpoint `/users/{userId}/orders/{orderId}` would make sense if we were using composite primary keys (e.g. an order is identified by `(userId, orderId)`, think of how GitHub repositories are identified), but our current implementation does not support composite keys AFAIK.

> Will make an issue. 

If the issue has been created, could you please cross-link that issue with this pull request

src/controllers/user-order.controller.ts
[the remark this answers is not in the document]
virkt25: See: https://github.com/strongloop/loopback-next/issues/1758

test/acceptance/shopping-cart.controller.acceptance.ts
```diff
@@ -3,34 +3,23 @@
...
 import {ShoppingApplication} from '../..';
 import {ShoppingCartRepository} from '../../src/repositories';
 import {RedisDataSource} from '../../src/datasources';
 import {ShoppingCart, ShoppingCartItem} from '../../src/models';
+import {setupApplication} from './helper';
 
 describe('ShoppingCartController', () => {
   let app: ShoppingApplication;
-  let server: RestServer;
   let client: supertest.SuperTest<supertest.Test>;
```
bajtos: Please use `Client` instead of `supertest.SuperTest<supertest.Test>`.

test/acceptance/user-order.controller.acceptance.ts
```diff
@@ -0,0 +1,156 @@
...
+import {supertest, expect} from '@loopback/testlab';
+import {ShoppingApplication} from '../..';
+import {OrderRepository, UserRepository} from '../../src/repositories';
+import {MongoDataSource} from '../../src/datasources';
+import {User, Order} from '../../src/models';
+import {setupApplication} from './helper';
+
+describe('UserOrderController acceptance tests', () => {
+  let app: ShoppingApplication;
+  let client: supertest.SuperTest<supertest.Test>;
```
bajtos: Ditto, use `Client` instead of `supertest.SuperTest<supertest.Test>`.

raymondfeng: I fixed that with #16

PR #13, opened 2018-09-17
feat: add an api to add items to a cart atomically

raymondfeng:
The PR adds an `addItem` api to atomically add items to a cart.

The implementation utilizes `execute`, which is a bit hacky.


src/repositories/shopping-cart.repository.ts
```diff
@@ -4,14 +4,36 @@
...
+
+  /**
+   * Add an item to the shopping cart with optimistic lock
+   * @param userId User id
+   * @param item Item to be added
+   */
+  async addItem(userId: string, item: ShoppingCartItem) {
+    const connector = this.kvModelClass.dataSource!.connector!;
+    // tslint:disable-next-line:no-any
+    const execute = promisify((cmd: string, args: any[], cb: Function) => {
```
virkt25: I'm not sure I follow what `execute` does? Apart for this and it's usage, the rest of it looks good.

src/models/shopping-cart.model.ts
```diff
@@ -25,6 +25,10 @@ export class ShoppingCartItem {
    */
   @property()
   price?: number;
+
+  constructor(data?: Partial<ShoppingCartItem>) {
```
virkt25: Is the partial needed in this case since this isn't a model directly exposed over `REST` and won't face issues of trying to create an instance without a `productId`?

raymondfeng: Not a big deal as the `data` itself is optional. So we can create a empty instance and populate the data afterward. The `required` constraint should be enforced by model validation.

src/repositories/shopping-cart.repository.ts
virkt25: But we're still using `this.get`, `this.set`? So what do the commands do?

raymondfeng: this.get/this.set delegate to execute behind the scene.

raymondfeng: The reason to use `execute` is that we don't have a `this.checkAndSet`. Maybe we should add it to the KVRepository, such as:
```ts
checkAndSet(key, current => {...});
```

raymondfeng: I'm considering the multiple device case. A user can access the app from multiple devices that share the same cart.

virkt25: But each call does a `get` and then `set`. Is a user really going to generate multiple requests at the same time from multiple devices? 

Either way, I'm ok with the change -- can you just add a comment stating the reason for using a transaction so it's clear to anyone looking at this example.

raymondfeng: Added more comments.

src/models/shopping-cart.model.ts
bajtos: > The required constraint should be enforced by model validation.

@raymondfeng Please note that KeyValue repository does not perform validation right now! See [lib/kvao/set.js](https://github.com/strongloop/loopback-datasource-juggler/blob/fcac276955a58c4aabcd2b602af046f45cfc73fe/lib/kvao/set.js). I think this is an oversight we should eventually fix.

src/repositories/shopping-cart.repository.ts
```diff
@@ -4,14 +4,69 @@
...
+      cart.items = cart.items || [];
+      cart.items.push(item);
+      return cart;
+    };
+    return this.checkAndSet(userId, addItemToCart);
+  }
+
+  /**
+   * Use Redis WATCH and Transaction to check and set against a key
+   * See https://redis.io/topics/transactions#optimistic-locking-using-check-and-set
```
bajtos: @raymondfeng IIUC, the algorithm outlined in the linked page is expecting to re-run the transaction if a race condition was detected, and repeat those re-run until the transaction succeed.

I don't see that logic implemented here. Are we expecting the REST clients to retry the operation? Could you please explain how are we signaling the race condition to them?

test/acceptance/shopping-cart.controller.acceptance.ts
```diff
@@ -102,12 +121,20 @@ describe('ShoppingCartController', () => {
     return new ShoppingCart({
       userId: 'user-0001',
       items: [
-        {
+        new ShoppingCartItem({
```
bajtos: Can you use `givenAnItem()` please?

PR #14, opened 2018-09-20
feat: enable retry for addItem and add more debug info

raymondfeng:
Follow up #13 per @bajtos comments:

- Enable `retry` for adding items to a shopping cart 
- Add more debug statements


src/controllers/shopping-cart.controller.ts
```diff
@@ -78,6 +84,64 @@ export class ShoppingCartController {
...
   }
 }
+
+/**
+ * Retry a task for number of times with the given interval in ms
+ * @param task Task object {run, description}
+ * @param maxRetries Maximum number of tries, default to 10
+ * @param interval Milliseconds to wait after each try, default to 100ms
+ */
+async function retry<T>(
```
b-admike: do we want tests for this function?

package-lock.json
```diff
@@ -100,6 +100,17 @@
         "globals": "^11.1.0",
         "invariant": "^2.2.0",
         "lodash": "^4.17.5"
+      },
```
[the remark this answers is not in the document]
raymondfeng: Yes, `package.lock` is good for an application to make it repeatable. For example, we can use [npm ci](https://docs.npmjs.com/cli/ci) and `npm audit fix`.

src/controllers/shopping-cart.controller.ts
```diff
@@ -78,6 +84,64 @@ export class ShoppingCartController {
...
+      maxRetries,
+    );
+    cart = await task.run();
+    if (cart != null) return cart;
+    if (--triesLeft > 0) {
+      debug('Wait for %d ms', interval);
+      await sleep(interval);
+    } else {
+      // No more retries, timeout
+      const msg = `Fail to ${task.description} after ${maxRetries *
```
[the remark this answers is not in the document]
raymondfeng: I'm fine with either tense.

src/controllers/shopping-cart.controller.ts
raymondfeng: We could add a unit test but not sure how to test it with redis.

raymondfeng: Added tests.

src/controllers/shopping-cart.controller.ts
```diff
@@ -78,6 +85,14 @@ export class ShoppingCartController {
...
   ) {
-    await this.shoppingCartRepository.addItem(userId, item);
+    debug('Add item %j to shopping cart %s', item, userId);
+    return retry<ShoppingCart>(
+      {
+        run: () => this.shoppingCartRepository.addItem(userId, item),
+        description: `update the shopping cart for '${userId}'`,
+      },
+      10,
+      100,
```
bajtos: I find these positional arguments difficult to understand. What is `10` and `100`?

I am proposing to change `retry` API to accept an options object instead of positional parameters.

```ts
    return retry<ShoppingCart>(
       {
         run: () => this.shoppingCartRepository.addItem(userId, item),
         description: `update the shopping cart for '${userId}'`,
       },
       {
         maxRetries: 10,
         retryInterval: 100, 
       });
```

src/repositories/shopping-cart.repository.ts
```diff
@@ -42,6 +42,9 @@ export class ShoppingCartRepository extends DefaultKeyValueRepository<
    * @param userId User id
    * @param check A function that checks the current value and produces a new
    * value. It returns `null` to abort.
+   *
+   * @returns A promise of the updated ShoppingCart instance or `null` if the
```
bajtos: I find this sentence ambiguous. If the transaction fails, is the function returning `null` or a Promise that's resolved to `null`?

Also: A transaction can fail for other reasons besides a race condition, for example the server can become unreachable or unresponsive. We are not returning `null` but rejecting the promise in those cases!

Proposed wording:

_A promise that's resolved with the updated ShoppingCart instance or with `null` if the transaction failed because of a race condition._

src/utils/retry.ts
```diff
@@ -0,0 +1,80 @@
...
+import * as debugFactory from 'debug';
+const debug = debugFactory('loopback:example:shopping');
+
+/**
+ * A task that can be retried
+ */
+export interface Task<T> {
+  run(): Promise<T | null | undefined>;
+  // tslint:disable-next-line:no-any
+  isDone?(err: any, result: T | null | undefined): result is T;
```
bajtos: [KISS](https://en.wikipedia.org/wiki/KISS_principle) & [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

Let's remove `isDone` from this pull request since we don't need it (yet).

src/controllers/shopping-cart.controller.ts
```diff
@@ -78,6 +85,14 @@ export class ShoppingCartController {
     @param.path.string('userId') userId: string,
     @requestBody({description: 'shopping cart item'}) item: ShoppingCartItem,
   ) {
-    await this.shoppingCartRepository.addItem(userId, item);
+    debug('Add item %j to shopping cart %s', item, userId);
+    return retry<ShoppingCart>(
```
bajtos: IMO, it should be the responsibility of a repository to deal with the need to retry this transaction. To me, it's an implementation detail specific to Redis storage. If we use a different backend that supports atomic partial updates, then `retry` won't be necessary.

It feels very wrong to me that we have to change controller code when the backing database is changed. The repository should be the part of the application dealing with database complexities.

Please move `retry` to `ShoppingCartRepository`.

src/utils/retry.ts
```diff
@@ -0,0 +1,80 @@
...
+        interval} ms`;
+      debug('%s', msg);
+      throw new HttpErrors.RequestTimeout(msg);
+    }
+  }
+}
+
+// tslint:disable-next-line:no-any
+function taskFinished<T>(err: any, result: T | null | undefined): result is T {
+  return err == null && result != null;
```
bajtos: Assuming we want to make `retry` generic: I think it's not safe to assume that `null`/`undefined` is always an invalid result signaling a need to retry.

I think we should remove `result != null` check and modify the shopping-cart code calling `retry` to provide `isDone` function too.

Alternatively, it may be better & easier to combine these two function into one and change the return value into an object with two properties: `done: boolean` and `value: T`.

- `{done: true, value: /*result*/}` - the task finished successfully
- `{done: false /* value can be omitted */}` - a retry is needed

One more important point: your current implementation retries regardless of whether the transaction was successfully executed with rollback result or whether there was an error which may prevent retries from succeeding (the commands are not valid, the server is not responding, etc.). That does not feel right to me, as it can retry a request that's known to fail again, unnecessarily putting more pressure on the backend system.

IMO, this `retry` function should fail immediately when the task fails. Implementing a circuit-switcher behavior is out of scope of this work.

test/acceptance/shopping-cart.controller.acceptance.ts
```diff
@@ -117,24 +117,26 @@ describe('ShoppingCartController', () => {
     await cartRepo.deleteAll();
   }
 
+  function givenAnItem(item?: Partial<ShoppingCartItem>) {
+    return new ShoppingCartItem(
+      item || {
```
bajtos: Data builders typically use `Object.assign`, allowing users to leverage sensible defaults provided by the data builder and customize only the properties relevant to the test.

```ts
const defaultData = {
 // sensible property values
};
return new ShoppingCartItem(
  Object.assign(defaultData, item)
);
```

test/acceptance/shopping-cart.controller.acceptance.ts
```diff
@@ -117,24 +117,26 @@ describe('ShoppingCartController', () => {
...
+
   function givenShoppingCart() {
     return new ShoppingCart({
       userId: 'user-0001',
       items: [
-        new ShoppingCartItem({
+        givenAnItem({
           productId: 'iPhone XS Max',
           quantity: 1,
           price: 1200,
```
bajtos: Just use `givenAnItem()`, the specific property values are not important here.

src/utils/retry.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Sleep for the given milliseconds
+ * @param ms Number of milliseconds to wait
+ */
+export function sleep(ms: number) {
+  return promisify(setTimeout)(ms);
```
bajtos: Please move `promisify(setTimeout)` out of this function, so that we don't have to create a new function whenever sleep is called.

```ts
const setTimeoutAsync = promisify(setTimeout);

export function sleep(ms) {
  return setTimeoutAsync(ms);
}
```

I think this can be further simplified as follows:

```ts
export const sleep = promisify(setTimeout);
```

raymondfeng:
@bajtos Great feedback. PTAL.

src/utils/retry.ts
```diff
@@ -0,0 +1,77 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {HttpErrors} from '@loopback/rest';
+import {promisify} from 'util';
+import * as debugFactory from 'debug';
+const debug = debugFactory('loopback:example:shopping');
+
+export interface TaskStatus<T> {
+  done: boolean;
+  value?: T | null;
```
bajtos: Nitpick: this type allows the developer to provide combinations of done/value that are not valid:

 - `{done: true}` (Missing value. What should be the outcome of the operation?)
 - `{done: false, value: 'some data'}`

src/utils/retry.ts
```diff
@@ -0,0 +1,77 @@
...
+  let triesLeft = maxTries;
+  while (true) {
+    debug(
+      'Try %s (%d/%d)',
+      task.description,
+      maxTries - triesLeft + 1,
+      maxTries,
+    );
+    const status = await task.run();
+    if (status.done) return status.value!;
```
bajtos:
```ts
status.value!
```

Please don't! The typings are designed in such way that `task.run()` can return a value that's undefined.

If you want to preserve `TaskStatus` interface then you need to change `retry` to return `T | null` and push the responsibility of handling null/undefined result to callers of `retry`. This is not a great UX to me.

I think a better solution is to follow my proposal outlined in https://github.com/strongloop/loopback4-example-shopping/pull/14#discussion_r219851802, I believe it provides information to allow TypeScript understand that when `status.done` is true, then `status.value` must be defined.

```ts
export interface TaskFinished<T> {
  done: true;
  result: T;
}

export interface RetryNeeded {
  done: false;
}

export interface Task<T> {
  run(): Promise<TaskFinished<T> | RetryNeeded>;
  description?: string;
}
```

src/utils/retry.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (--triesLeft > 0) {
+      debug('Wait for %d ms', interval);
+      await sleep(interval);
+    } else {
+      // No more retries, timeout
+      const msg = `Failed to ${task.description} after ${maxTries *
+        interval} ms`;
+      debug('%s', msg);
+      throw new HttpErrors.RequestTimeout(msg);
```
bajtos: Nitpick: Since `retry` is not used at REST layer, but inside a repository, I think we shouldn't be throwing an HttpError and use `err.code` to provide a machine-readable flag that can be used to detect this particular error.

PR #16, opened 2018-09-24
feat: add a default home page

raymondfeng:
Add a default home page for the app. We'll use it as a template for https://github.com/strongloop/loopback-next/issues/1745.

We need to fix `@loopback/rest` so that it respects the correct `content-type` from `responses`. At the moment, the home page will render raw html as the media type is `plain/text` for string response from the controller method.


virkt25:
I made the following changes to get the page to render :D 

image


Render: 

image


Looks good overall -- have some concerns with the resources we're loading -- will leave a comment.

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+
+<head>
+  <title>${this.description}</title>
+
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+
+  <!-- Latest compiled and minified CSS -->
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u"
```
virkt25: Can we skip out on loading the full `bootstrap.min.css`  and the accompanying `js` file? We're using about 2 styles from this entire file. I'd rather we copy the style in-lined for a faster render. No JS is needed. (Currently this + JS is loading 63.1 KB while the page payload is only 1.7 KB). 

An alternative IIRC is that BootStrap also provides just the `grid` CSS -- which is lighter and includes the CSS for `container` class.

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+      <p>
+        Powered by <a href="https://v4.loopback.io">
+          <img src="http://v4.loopback.io/img/global/loopback-full-logo-blue.svg" width="100px" />
+        </a>
+      </p>
+    </div>
+  </footer>
+  <!-- Bootstrap core JavaScript -->
+  <!-- Placed at the end of the document so the pages load faster -->
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
```
virkt25: Nothing on this page needs the JS ... no BootStrap widgets are used.

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  <!-- Latest compiled and minified CSS -->
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u"
+    crossorigin="anonymous">
+</head>
+
+<body>
+  <div class="container">
+    <h1>${this.name}@${this.version}</h1>
+    <ul>
+      <li><a href="/openapi.json">OpenAPI spec: /openapi.json</a></li>
```
virkt25: Can we style the list to be bigger. Plus can `OpenAPI spec:` not be part of the link. Bolded. Would be a better style imo. 

Same applies for the bullet below.

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+      <li><a href="/explorer">API Explorer: /explorer</a></li>
+    </ul>
+  </div>
+  <footer>
+    <div class="container">
+      <p>
```
virkt25: Let's center this text on the page.

raymondfeng:
@virkt25 You worked around the issue by manipulating Express response directly. In this case, we should introduce a special return value such as `ByPassResponse` so that our writer won't try to send 204.

Feel free to improve the UI. I'm a beginner there :-).

virkt25:
> You worked around the issue by manipulating Express response directly. In this case, we should introduce a special return value such as ByPassResponse so that our writer won't try to send 204.

I think a simpler fix can even be just to check if `Content-Type` header has been set in the writer ... and if so we don't overwrite it -- and then we can rely on the `response.end` in the write.

> Feel free to improve the UI. I'm a beginner there :-).

I'll push some changes to the PR in a bit :)

raymondfeng:
Please the CI failure is not related to this change. It will be fixed as part of #14.

virkt25:
Here's a screenshot for the UI -- no external CSS / JS. Minor variances with some alternatives below. I personally vote for Alternative 1 (I tried it after pushing my changes otherwise I would've included that in the PR). Anyone else have any preferences? cc: @raymondfeng @bajtos @strongloop/sq-lb-apex 

Current:
image
---
Alternative 1: 
image
---
Alternative 2:
image
---
Alternative 3:
image
---

bajtos:
I vote for Alternative 1.

Perhaps instead of using `name@version` format, can we move the version number to line 2, to make it more explicit what is version and what is the app name? 

A mock-up:

<p align=center>
<strong>@loopback/example-shopping</strong>
<br/>
<em>version 1.0.0</em>
</p>

public/index.html.template
```diff
@@ -0,0 +1,50 @@
...
+      left: 50%;
+      transform: translateX(-50%);
+    }
+  </style>
+</head>
+
+<body>
+  <h1 style="text-align: center">${name}@${version}</h1>
+  <h3>OpenAPI spec: <a href="/openapi.json">/openapi.json</a></h4>
+  <h3>API Explorer: <a href="/explorer">/explorer</a></h4>
```
bajtos: IIRC, applications can customize the URL of both OpenAPI spec and Explorer. See https://github.com/strongloop/loopback-next/pull/1637, [Customize How OpenAPI Spec is Served](https://github.com/strongloop/loopback-next/blob/eca79a45244c3e9215face9b9a35b3d3d2e5cb0a/docs/site/Server.md#customize-how-openapi-spec-is-served) and [Configure the API Explorer](https://github.com/strongloop/loopback-next/blob/eca79a45244c3e9215face9b9a35b3d3d2e5cb0a/docs/site/Server.md#configure-the-api-explorer).

IMOL, the home page should respect these settings. Eventually :) I am fine to keep the current static version as the first baby step.

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const pkg = require('../../../package.json');
+
+export class HomePageController {
+  name: string;
+  version: string;
+  description: string;
+  render: TemplateExecutor;
+
+  constructor(@inject(RestBindings.Http.RESPONSE) private res: Response) {
```
bajtos: Can we use `response` instead of `res` please?

src/controllers/home-page.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {get} from '@loopback/openapi-v3';
+import * as fs from 'fs';
+import * as path from 'path';
+import {template, TemplateExecutor} from 'lodash';
+import {inject} from '@loopback/context';
+import {RestBindings, Response} from '@loopback/rest';
+
+const pkg = require('../../../package.json');
```
bajtos: From the perspective of clean design and Dependency Injection pattern, I think controller should not depend `package.json` directly, but get it injected from app context. Then the application should bind the content of `package.json` (maybe a subset of properties only?) in its constructor.

Not a big deal for this particular pull request, but I think we should think more about the design patterns when we start working on adding HomePageController to our project template.

raymondfeng: Yes, I'm thinking about the same direction while I work on this implementation.

Issue #15, opened 2018-09-23
Suggestion Models in their own files.


marioestradarosa:
To keep it standardized with other **LB4** tutorials and model discovery utility functions, you should place  the two models in their own file.

src/models/shopping-cart-item.model.ts
src/models/shopping-cart.model.ts

hacksparrow:
Agreed.

PR #18, opened 2018-09-25
chore: separate shopping cart item into its own file

raymondfeng:
Fixes https://github.com/strongloop/loopback4-example-shopping/issues/15

Signed-off-by: Raymond Feng <enjoyjava@gmail.com>

### DIFF
--- a/src/models/shopping-cart-item.model.ts
+++ b/src/models/shopping-cart-item.model.ts
@@ -1,0 +1,32 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-shopping
+// This file is licensed under the MIT License.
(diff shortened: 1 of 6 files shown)


src/models/shopping-cart-item.model.ts
```diff
@@ -0,0 +1,32 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+/**
+ * Item in a shopping cart
+ */
+@model()
+export class ShoppingCartItem extends Entity {
+  /**
```
b-admike: Nitpick: the comments for the properties seem redundant, so I suggest removing them, or adding some more meat to them.

PR #19, opened 2018-09-25
fix: update template

virkt25:
Signed-off-by: virkt25 <taranveer@virk.cc>

Update the home page template.

### DIFF
--- a/public/index.html.template
+++ b/public/index.html.template
@@ -7,6 +7,7 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="shortcut icon" type="image/x-icon" href="//v4.loopback.io/favicon.ico">
 
   <style>
     h3 {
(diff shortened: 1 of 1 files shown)


src/controllers/home-page.controller.ts
```diff
@@ -17,6 +17,7 @@ export class HomePageController {
   constructor(
     @inject(PackageKey) private pkg: PackageInfo,
     @inject(RestBindings.Http.RESPONSE) private response: Response,
+    @inject(RestBindings.HTTPS_OPTIONS) private config: RestServerConfig,
```
virkt25: So ... we are binding the RestConfig to `HTTPS_OPTIONS` ... and I tried to re-bind the config back to `RestBindings.CONFIG` but that threw an error since that key contains a `#` (it's accessing `ApplicationConfig#rest`). So not sure what we should change in `@loopback/rest` ... hopefully this starts the discussion around it.

raymondfeng: RestBindings.CONFIG cannot be used for binding as it's a property of `ApplicationConfig`.

raymondfeng:
Let's keep `/openapi.json` hard-coded now without dealing with customization. It's getting too complicated for the home page controller.

raymondfeng:
@virkt25 Can you have a PR for the style change itself, which is much less controversial?

virkt25:
@raymondfeng done.

PR #20, opened 2018-09-27
feat: product recommendation 

virkt25:
fixes https://github.com/strongloop/loopback-next/issues/1482

---

implements a mock recommendation service -- runs on port 3001.
implements a proxy service to talk to the recommendation service. 

both services start by running `npm run start`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,9 @@
     "posttest": "npm run lint",
     "test:dev": "lb-mocha --allow-console-logs DIST/test/**/*.js && npm run posttest",
     "prestart": "npm run build",
-    "start": "node .",
+    "start": "concurrently --kill-others \"npm run start:app\" \"npm run start:recommender\"",
+    "start:app": "node .",
+    "start:recommender": "node recommender",
     "prepublishOnly": "npm run test"
   },
   "repository": {
@@ -59,12 +61,15 @@
     "@loopback/openapi-v3": "^0.15.0",
     "@loopback/repository": "^0.18.0",
     "@loopback/rest": "^0.25.0",
+    "@loopback/service-proxy": "^0.8.6",
     "bcryptjs": "^2.4.3",
     "debug": "^4.0.1",
+    "express": "^4.16.3",
     "isemail": "^3.1.3",
     "lodash": "^4.17.11",
     "loopback-connector-kv-redis": "^3.0.0",
-    "loopback-connector-mongodb": "^3.8.0"
+    "loopback-connector-mongodb": "^3.8.0",
+    "loopback-connector-rest": "^3.3.0"
   },
   "devDependencies": {
     "@commitlint/cli": "^7.1.2",
@@ -74,10 +79,12 @@
     "@loopback/testlab": "^0.14.0",
     "@types/bcryptjs": "^2.4.2",
     "@types/debug": "^0.0.30",
+    "@types/express": "^4.16.0",
     "@types/lodash": "^4.14.116",
     "@types/mocha": "^5.0.0",
     "@types/node": "^10.11.0",
     "commitizen": "^2.10.1",
+    "concurrently": "^4.0.1",
     "cz-conventional-changelog": "^2.1.0",
     "husky": "^0.14.3",
     "mocha": "^5.2.0",
--- a/recommender.js
+++ b/recommender.js
@@ -12,4 +12,4 @@ const app = require(path.join(
   '/recommender',
 ));
 
-module.exports = app.recommender.start();
+module.exports = app.createRecommendationServer().start();
--- a/recommender.js
+++ b/recommender.js
@@ -1,0 +1,15 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+const distUtil = require('@loopback/dist-util');
+const path = require('path');
+
+const app = require(path.join(
+  __dirname,
+  distUtil.getDist(__dirname),
+  '/recommender',
+));
+
+module.exports = app.recommender.start();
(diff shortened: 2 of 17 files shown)


src/services/recommender.service.ts
```diff
@@ -0,0 +1,25 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {getService} from '@loopback/service-proxy';
+import {inject, Provider} from '@loopback/core';
+import {RecommenderDataSource} from '../datasources/recommender.datasource';
+import {Product} from '../models';
+
+export interface RecommenderService {
+  recommend(userId: string): Promise<Product[]>;
```
raymondfeng: `getProductRecommendations()`?

tsconfig.json
```diff
@@ -4,6 +4,7 @@
   "include": [
     "src",
     "test",
+    "recommender",
```
raymondfeng: At some point, we probably need to convert this repo to be `lerna` managed monorepo.

recommender/mock-recommendation-app.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  app.get('/:userId', (req: express.Request, res: express.Response) => {
+    res.send(recommendations);
+  });
+
+  return {
+    start: () => {
+      server = app.listen(3001, () => {
+        console.log(
+          'Mock Product Recommender powered by Express started on port 3001',
```
raymondfeng: Use `${port}` instead of hard-coded 3001.

recommender/mock-recommendation-app.ts
```diff
@@ -0,0 +1,31 @@
...
+  let server: Server;
+  const app = express();
+
+  app.get('/:userId', (req: express.Request, res: express.Response) => {
+    res.send(recommendations);
+  });
+
+  return {
+    start: () => {
+      server = app.listen(3001, () => {
```
raymondfeng: Use `port`

recommender/mock-recommendation-app.ts
```diff
@@ -0,0 +1,31 @@
...
+        console.log(
+          'Mock Product Recommender powered by Express started on port 3001',
+        );
+      });
```
raymondfeng: Should it return `server`?

recommender.js
```diff
@@ -0,0 +1,15 @@
...
+const distUtil = require('@loopback/dist-util');
+const path = require('path');
+
+const app = require(path.join(
+  __dirname,
+  distUtil.getDist(__dirname),
+  '/recommender',
+));
+
+module.exports = app.createRecommendationServer().start();
```
raymondfeng: I think it should export `app.createRecommendationServer()` instead so that it can be stopped`.

package.json
```diff
@@ -36,7 +36,9 @@
     "posttest": "npm run lint",
     "test:dev": "lb-mocha --allow-console-logs DIST/test/**/*.js && npm run posttest",
     "prestart": "npm run build",
-    "start": "node .",
+    "start": "concurrently --kill-others \"npm run start:app\" \"npm run start:recommender\"",
```
raymondfeng: Why not `npm run start:recommender && npm run start:app`?

recommender/mock-recommendation-app.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import * as express from 'express';
+import {Server} from 'http';
+const recommendations = require('./recommendations.json');
+
+export function createRecommendationServer(port: number = 3001) {
```
raymondfeng: It should just take `(options: http.ServerOptions = {port: 3001})`

package.json
virkt25: Because then the start process hangs at `npm run start:recommeder` as the server is running ad doesn't return a response to the command line to allow it to continue to the next `npm` script. 

We used `concurrently` in https://github.com/strongloop/loopback4-example-microservices as well.

recommender/mock-recommendation-app.ts
virkt25: No other options are configurable in the app and I can't find a `http.ServerOptions` type.

package.json
raymondfeng: Another option: https://www.npmjs.com/package/npm-run-all

recommender/mock-recommendation-app.ts
raymondfeng: Oh, we only have `https.ServerOptions`.

raymondfeng: Keeping it an `options` object will allow us to configure it to be https too.

recommender/mock-recommendation-app.ts
```diff
@@ -0,0 +1,21 @@
...
+  const app = express();
+
+  app.get('/:userId', (req: express.Request, res: express.Response) => {
+    res.send(recommendations);
+  });
+
+  return app.listen(port, () => {
+    console.log(
+      'Mock Product Recommender powered by Express started on port 3001',
```
raymondfeng: Please use `${port}.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -107,6 +110,23 @@ describe('UserController', () => {
...
+    before(() => {
+      recommendationService = createRecommendationServer();
+    });
+
+    after(() => {
+      recommendationService.close();
+    });
+
+    it('returns product recommendations for a user', async () => {
+      await client.get(`/users/userid/recommend`).expect(200, recommendations);
```
b-admike: I'm probably missing something, but shouldn't `userid` be a number here?

virkt25: Mongo generated UUID for User model.

recommender/mock-recommendation-app.ts
virkt25: So I should set the type to be `https.ServerOptions`?

PR #21, opened 2018-09-28
Add overview diagram

dhmlau:
Per @raymondfeng's comment, we should have an overview diagram for this shopping example. 

Signed-off-by: Diana Lau <dhmlau@ca.ibm.com>


raymondfeng:
@dhmlau Please fix the commit message.

README.md
```diff
@@ -6,6 +6,8 @@ This project aims to represent an online ecommerce platform APIs to validate /
 test the LoopBack 4 framework readiness for GA. See
 https://github.com/strongloop/loopback-next/issues/1476 for more information.
 
+![Shopping example overview diagram](example-shopping.png)
```
virkt25: The URLs aren't really reflective of the full set of REST APIs we expose.

dhmlau:
@virkt25 , fixed.

raymondfeng:
@dhmlau I actually like the 1st version of the diagram layout better.

dhmlau:
@raymondfeng @virkt25, here is the original diagram: 
screen shot 2018-09-30 at 3 23 24 pm

In a previous comment from @virkt25 , he mentioned that the endpoints shown in the diagram above does not list all the available endpoints, so it may be better to have it removed.  Below shows the endpoints we have: 
explorer

I have 3 proposals: 
1. shows all the endpoints: /shoppingCarts/{userId} & /shoppingCars/{userId}/items, etc. 
   - but this may not be accurate because not all the methods support that endpoint.

2. Only shows the base url, i.e. /shoppingCarts, /users/, /users/{userId}/orders
   - Order history one is a bit odd, but i guess it's ok because it's "belongs to" a user.  

3. Revert back to the original diagram.
   - IMHO, it seems to be a better way if we want to show the endpoints. 

@strongloop/sq-lb-apex @raymondfeng @bajtos, what do you think?

b-admike:
@dhmlau I think the original diagram is already good, but from the three proposals you made, my choices would be 2, 3, and 1 in that order.

PR #23, opened 2018-11-02
chore: upgrade to LoopBack4@1.x

raymondfeng:
Upgrade to LB4@1.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,12 +11,9 @@
     "node": ">=8"
   },
   "scripts": {
-    "build:all-dist": "npm run build:dist8 && npm run build:dist10",
     "build:apidocs": "lb-apidocs",
-    "build": "lb-tsc",
-    "build:dist8": "lb-tsc es2017",
-    "build:dist10": "lb-tsc es2018",
-    "build:watch": "lb-tsc --watch",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -11,12 +11,9 @@
...
   },
   "scripts": {
-    "build:all-dist": "npm run build:dist8 && npm run build:dist10",
     "build:apidocs": "lb-apidocs",
-    "build": "lb-tsc",
-    "build:dist8": "lb-tsc es2017",
-    "build:dist10": "lb-tsc es2018",
-    "build:watch": "lb-tsc --watch",
+    "build": "lb-tsc es2017 --outDir dist --copy-resources",
+    "build:watch": "lb-tsc es2017 --outDir dist --watch",
```
jannyHou: any reason remove the node 10 build script `"build:dist10": "lb-tsc es2018"` ?

jannyHou: hmm, any reason downgrade the es2018 to es2017 for node10?
according to https://node.green/#ES2018 I think es2018 is still a good target for node 10.

raymondfeng: This is to be consistent with `loopback-next`. As we decide to have one distribution, we need to pick the lower denominator.

PR #26, opened 2018-12-11
feat: jwt auth

jannyHou:
*I am going to fix the CI. It wouldn't block the review, the tests all pass on my local*

```


  35 passing (1s)
  3 pending
```

connect to https://github.com/strongloop/loopback-next/issues/1997

This PR implements the JWT authentication strategy for verifying requests:

- A JWT strategy class
  - includes an `authenticate` method that retrieves the token from request and then restores the user profile by decoding the token.
- An auth action provider
  - executes the `jwtStrategy.authenticate()` function.
- A strategy resolver provider
  - returns a JWTStrategy instance if an endpoint is decorated with `@authentication('jwt')`
- Two endpoints in the user controller
  - users/login
    - jwt token will be generated using user's email and id as payload
  - users/me
    - returns the login user


src/controllers/user.controller.ts
```diff
@@ -91,7 +93,7 @@ export class UserController {
   }
 
   @post('/users/login')
-  async login(@requestBody() user: User): Promise<User> {
+  async login(@requestBody() user: User): Promise<string | undefined> {
```
jannyHou: We need a type filter to describe the credentials needed for login.

The filtered type would be

```ts
type Credentials = {
   username: string,
  // consider adding `Password` and `Email` as built-in loopback types
   password: Password,
   email: Email
}
```

Or 

```ts
type Credentials = {
  identity: string | Email
  password: Password
}
```

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (token) {
+      try {
+        return await verifyAsync(token, SECRET);
+      } catch (err) {
+        if (err) return Promise.reject('Authentication failed!');
+      }
+    }
+
+    return Promise.reject('Token not found!');
```
bajtos: When the client did not provide any token, perhaps because it did not know this URL requires authentication, the server should return `401 Unauthorized` response with `WWW-Authenticate` header field.

In Passport, the strategy would call `this.fail()` API, see e.g. how `passport-jwt` uses it here: https://github.com/themikenicholson/passport-jwt/blob/f2044371bc9c85f22fbd30e8e58267a42dcd26ac/lib/strategy.js#L96

In `@loopback/authentication`, the strategy adapter is implementing `fails` as follows. I think this is not entirely correct implementation, e.g. I don't see any `WWW-Authenticate` header field being set.

https://github.com/strongloop/loopback-next/blob/aacfcf7b1ee8f6550b91d1fa3c84a6e4fbfc9a36/packages/authentication/src/strategy-adapter.ts#L50-L53
```ts
      // add failure state handler to strategy instance
      strategy.fail = function(challenge: string) {
        reject(new HttpErrors.Unauthorized(challenge));
      };
```

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+import {promisify} from 'util';
+const verifyAsync = promisify(jwt.verify);
+// Consider turn it to a binding
+const SECRET = 'secretforjwt';
+import {Request, HttpErrors} from '@loopback/rest';
+import {UserProfile} from '@loopback/authentication';
+import * as _ from 'lodash';
+
+export class JWTStrategy {
+  // tslint:disable-next-line:no-any
```
bajtos: I think this tslint directive is no longer needed, can we remove it?

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    // there is a discussion regarding how to retrieve the token,
+    // see comment https://github.com/strongloop/loopback-next/issues/1997#issuecomment-451054806
+    const token = request.query.token || request.headers['authorization'];
+    if (token) {
+      try {
+        const decoded = await verifyAsync(token, SECRET);
+        return Promise.resolve(_.pick(decoded, ['id', 'email']));
```
bajtos: I think you should pick `name` too, since that's a property described in UserProfileSchema below?

src/controllers/user.controller.ts
```diff
@@ -87,4 +160,56 @@ export class UserController {
...
+    if (credentials.password.length < 8) {
+      throw new HttpErrors.UnprocessableEntity(
+        'password must be minimum 8 characters',
+      );
+    }
+
+    // Check if user exists
+    try {
+      const foundUser = await this.userRepository.findOne({
+        where: {email: credentials.email, password: credentials.password},
```
bajtos: Uff uff, storing password in plain text is a huge security vulnerability. I recommend to use `scrypt` algorithm and store the hash only. Note that `scrypt` is built into Node.js from version 10, see https://nodejs.org/dist/latest-v10.x/docs/api/crypto.html#crypto_crypto_scrypt_password_salt_keylen_options_callback Since we are still supporting Node.js 8, we cannot use that built-in API and have to use a 3rd party module like https://www.npmjs.com/package/scrypt.

Historically (LoopBack 1.x/2x./3.x), we were using `bcrypt` instead.

src/controllers/user.controller.ts
```diff
@@ -87,4 +160,56 @@ export class UserController {
...
+      });
+      let token = undefined;
+      if (foundUser) {
+        const currentUser = _.pick(foundUser.toJSON(), ['id', 'email']);
```
bajtos: How about the `name` property?

src/controllers/user.controller.ts
```diff
@@ -87,4 +160,56 @@ export class UserController {
...
+        // Generate user token using JWT
+        token = await signAsync(currentUser, 'secretforjwt', {
+          expiresIn: 300,
+        });
+      } else {
+        return Promise.reject(
```
bajtos: Just `throw`.

src/providers/custom.authentication.provider.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(
+    // The provider is instantiated for Sequence constructor,
+    // at which time we don't have information about the current
+    // route yet. This information is needed to determine
+    // what auth strategy should be used.
+    // To solve this, we are injecting a getter function that will
+    // defer resolution of the strategy until authenticate() action
+    // is executed.
+    @inject.getter(AuthenticationBindings.STRATEGY)
+    readonly getStrategy: Getter<JWTStrategy>,
```
bajtos: What if the application uses multiple strategies? Can we have a more generic interface (e.g. `AuthStrategy`) instead of `JWTStrategy`?

src/repositories/user.repository.ts
```diff
@@ -18,13 +18,14 @@ export class UserRepository extends DefaultCrudRepository<
   typeof User.prototype.id
 > {
   public orders: HasManyRepositoryFactory<Order, typeof User.prototype.id>;
+  // From @jannyhou: should the relation here be hasOne or hasMany?
```
bajtos: I think this comment is no longer relevant, can you remove it?

src/controllers/user.controller.ts
```diff
@@ -87,4 +160,56 @@ export class UserController {
...
+          },
+        },
+      },
+    },
+  })
+  async login(
+    @requestBody() credentials: Credentials,
+  ): Promise<String | undefined> {
+    // Validate Email
+    if (!isemail.validate(credentials.email)) {
```
bajtos: Please move the implementation of `login` into `UserRepository` class, so that new tokens can be created from TS/JS API directly (e.g. in integration tests), without the need to go through the REST API.

src/types/custom-types.ts
```diff
@@ -0,0 +1,4 @@
+export type Credentials = {
```
bajtos: Please move this type definition to `user.repository.ts`.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -130,6 +133,74 @@ describe('UserController', () => {
...
+    it('login returns a valid token', async () => {
+      const newUser = await userRepo.create(user);
+      await client
+        .post('/users/login')
+        .send({email: newUser.email, password: newUser.password})
+        .expect(200)
+        .then(getToken);
+
+      function getToken(res: Response) {
+        // is it good to return the token as text?
```
bajtos: Not really. Please return a data object serialized as JSON, pass the token in a property of that object. For example:

```json
{
  "token": "header.payload.signature"
}
```

test/acceptance/user.controller.acceptance.ts
```diff
@@ -130,6 +133,74 @@ describe('UserController', () => {
...
+        .post('/users/login')
+        .send({email: newUser.email, password: newUser.password})
+        .expect(401);
+    });
+
+    it('/me returns the current user', async () => {
+      const newUser = await userRepo.create(user);
+      let token = '';
+      await client
+        .post('/users/login')
```
bajtos: Please create the token via UserRepository API, no need to make a (slow) HTTP request here.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -130,6 +133,74 @@ describe('UserController', () => {
...
+        .expect(200)
+        .then(getToken);
+
+      function getToken(res: Response) {
+        // is it good to return the token as text?
+        token = res.text;
+      }
+      // MongoDB returns an id object we need to convert to string
+      // since the REST API returns a string for the id property.
+      newUser.id = newUser.id.toString();
```
bajtos: Have you considered using `toJSON` helper from `@loopback/testlab` instead? See https://github.com/strongloop/loopback-next/tree/master/packages/testlab#toJSON

```
const me = _.pick(toJSON(newUser), ['id', 'email']));
```

test/acceptance/user.controller.acceptance.ts
```diff
@@ -130,6 +133,74 @@ describe('UserController', () => {
...
+
+      await client
+        .get('/users/me')
+        .set('Authorization', token)
+        .expect(200, me);
+    });
+
+    it('/me returns 401 when the token is not provided', async () => {
+      const newUser = await userRepo.create(user);
+      // tslint:disable-next-line:no-unused-variable
```
bajtos: This directive should not be needed, because the token is used below, right?

test/acceptance/user.controller.acceptance.ts
```diff
@@ -130,6 +133,74 @@ describe('UserController', () => {
...
+      let token = '';
+      await client
+        .post('/users/login')
+        .send({email: newUser.email, password: newUser.password})
+        .expect(200)
+        .then(getToken);
```
bajtos: Use `await` instead of `then`.

```ts
const res = await client
    .post('/users/login')
    .send({email: newUser.email, password: newUser.password})
    .expect(200);
const token = res.text;
```

Even better, rewrite this code to create the token by calling UserRepository API directly, no need to make HTTP calls here.

src/authentication-strategies/JWT.strategy.ts
jannyHou: @bajtos The picked properties are a little bit tricky.

- The type `UserProfile` only has `id` as required, which makes sense. `email` and `name` are optional properties.
- The `User` model doesn't have a `name` property but has `firstName` and `lastName` instead, we could compose a `name` property by joining them, but given the fact that `name` is optional, is it necessary to compose one?

test/acceptance/user.controller.acceptance.ts
jannyHou: @bajtos The tests are in the acceptance test file, that's why they make HTTP calls, if you look at the other test cases in this file they do the same :)

src/authentication-strategies/JWT.strategy.ts
bajtos: This is tricky indeed.

I think we should make `UserProfile` more flexible to allow the applications to choose what fields they want to include. That's probably out of scope of this spike though.

I would assume that UserProfile's `name` property is meant either as a handle or a display name. If we have firstName and lastName properties in our domain model, then I propose to use `firstName` as the profile `name`.

@raymondfeng what do you think?

test/acceptance/user.controller.acceptance.ts
bajtos: IMO, the tests should make HTTP calls only for things they are testing. For example, a test for `login` should call login via HTTP. Other tests should use a token created directly via JS/TS API, because token creation is not really a part of the tested scenario.

There are two reasons why I am advocating for this approach:
- Calling Repository API is simpler, it makes the code easier to read, and also tends to produce more helpful error message if something goes wrong.
- Calling Repository API directly makes the tests run faster. HTTP requests add a small overhead that adds up when there are many test cases to run.

src/authentication-strategies/JWT.strategy.ts
jannyHou: > I would assume that UserProfile's name property is meant either as a handle or a display name. If we have firstName and lastName properties in our domain model, then I propose to use firstName as the profile name.

I am good with doing it in this PR, and for the discussion about the flexibility, I created a new issue for it, let's continue the discussion there: https://github.com/strongloop/loopback-next/issues/2246

src/controllers/user.controller.ts
nabdelgadir: We're returning because we want a `401 Unauthorized Error` to be returned when someone logs in with the wrong credentials instead of receiving a generic `Error`.

bajtos: As I see it, the problem is in the way how this code is structured, in particular with this try/catch block:

```ts
try {
  // do everything
  // maybe return Promise.reject()
} catch (err) {
  return Promise.reject(new Error('Wrong Credentials!'));
}
```

I am sure you can find a way how to restructure this block so that we can use `throw` instead of `Promise.reject` and `Promise.resolve`.

test/acceptance/user.controller.acceptance.ts
jannyHou: neat 👍 rewrote the test to leverage repository APIs

src/controllers/user.controller.ts
```diff
@@ -102,6 +100,7 @@ export class UserController {
   async printCurrentUser(
     @inject('authentication.currentUser') currentUser: UserProfile,
   ): Promise<UserProfile> {
+    console.log(currentUser);
```
bajtos: Forgotten `console.log`? Maybe you should use `debug` instead?

src/repositories/user.repository.ts
```diff
@@ -30,4 +41,41 @@ export class UserRepository extends DefaultCrudRepository<
       async () => orderRepository,
     );
   }
+
+  async login(credentials: Credentials) {
+    // Validate Email
+    if (!isemail.validate(credentials.email)) {
+      throw new HttpErrors.UnprocessableEntity('invalid email');
```
bajtos: Ah, I see there is a catch in what I proposed.

The Repository layer should not be coupled with HTTP/REST specifics. Instead of using `HttpErrors.UnprocessableEntity` and signal the particular error case via `statusCode` property, we should use just `err.code` here and have this code translated to transport specific means elsewhere. For example, gRPC may want to use a different mechanism to signal malformed request payload.

You can find the current translation table in `@loopback/rest` here:

https://github.com/strongloop/loopback-next/blob/e0f75ace756d8b5479ad5b66e9b2803e4b419eef/packages/rest/src/providers/reject.provider.ts#L12-L16
```ts
// TODO(bajtos) Make this mapping configurable at RestServer level,
// allow apps and extensions to contribute additional mappings.
const codeToStatusCodeMap: {[key: string]: number} = {
  ENTITY_NOT_FOUND: 404,
};
```

Unfortunately, we don't yet implemented API allowing extensions to contribute custom mappings. It's something we want to eventually provide, but there was no urgent need so far.

I am proposing the following approach here in this spike:
1. Make sure that all authorization-related errors come with an error code, e.g. `UNPROCESSABLE_ENTITY`.
2. For this spike, set `statusCode` too (by calling one of `HttpErrors` factories)
3. Add a comment to remind us that this solution is only temporary, and we need to make `codeToStatusCodeMap` customizable by extensions. Create a user story for this task, it will become a part of this spike's outcome.

See also https://github.com/strongloop/loopback-next/commit/7a56bad1430e7258bba137ce24d4fdaecdb394fd, https://github.com/strongloop/loopback-next/pull/1658 and https://github.com/strongloop/loopback-next/issues/1469

src/repositories/user.repository.ts
```diff
@@ -30,4 +41,41 @@ export class UserRepository extends DefaultCrudRepository<
...
+
+      let token = undefined;
+      if (foundUser) {
+        const currentUser = _.pick(toJSON(foundUser), ['id', 'email']);
+        // Generate user token using JWT
+        token = await signAsync(currentUser, 'secretforjwt', {
+          expiresIn: 300,
+        });
+      } else {
+        return Promise.reject(
```
bajtos: Let's rewrite this block as follows - the code below is just an illustration, may not work OOTB.

```ts
const foundUser = await this.findOne({
  where: {email: credentials.email, password: credentials.password},
});

if (!foundUser) {
  new HttpErrors.Unauthorized('Wrong credentials!')
}

const currentUser = _.pick(toJSON(foundUser), ['id', 'email']);

// Generate user token using JWT
token = await signAsync(currentUser, 'secretforjwt', {
  expiresIn: 300,
});

return token;
```

See the next two comments for explanation.

src/repositories/user.repository.ts
```diff
@@ -30,4 +41,41 @@ export class UserRepository extends DefaultCrudRepository<
...
+          new HttpErrors.Unauthorized('Wrong credentials!'),
+        );
+      }
+      return Promise.resolve(token);
```
bajtos: No need to wrap the value in `Promise.resolve`, async functions always return promises. Just `return token`.

src/repositories/user.repository.ts
```diff
@@ -30,4 +41,41 @@ export class UserRepository extends DefaultCrudRepository<
...
+    } catch (err) {
+      return Promise.reject(new Error('Wrong Credentials!'));
```
bajtos: This code is throwing away all details of the original error (`err`), including the error message and stack trace, and reporting a generic error with no details. This is an anti-pattern, such code makes it very difficult to troubleshoot errors. Especially if they happen only in production on every other Friday night.

Async functions forward any unhandled errors as rejected promises, no extra code is needed here.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -184,18 +185,11 @@ describe('UserController', () => {
 
     it('/me returns 401 when the token is not provided', async () => {
       const newUser = await userRepo.create(user);
-      // tslint:disable-next-line:no-unused-variable
-      let token = '';
+
       await client
         .post('/users/login')
```
bajtos: Can we call Repository API directly here, bypassing the REST layer?

test/acceptance/user.controller.acceptance.ts
bajtos: > rewrote the test to leverage repository APIs

@jannyHou I don't see those changes in this commit. AFAICT, login is still invoked via `await client.post('/users/login'). Am I reading the diff incorrectly?

jannyHou: @bajtos latency 🙈 we just pushed the new commit, maybe refresh again?

src/controllers/user.controller.ts
jannyHou: @bajtos We changed the login API to return the token in a json payload. 

>  I recommend to use scrypt algorithm and store the hash only. Note that scrypt is built into Node.js from version 10,

Hmm, we didn't include the password in the payload that's used to generate the token, so I think we don't need `scrypt` here. 

Or did you mean we shouldn't store the password directly in the database but the hash of it instead? That's probably out of the scope of this PR, but good point, and I can create another PR to fix it.

src/repositories/user.repository.ts
jannyHou: @bajtos I found a really weird thing, the `Error` interface in `@types/node` doesn't have the `code` property, it only has an optional property `stack`. Which means by `let err = new Error('Wrong Credentials!')`, you cannot give `err` an error code. I could define a custom Error class that extend `Error` but if having to do that, I would rather leveraging the `HttpErrors` class from `rest` to create the error instance.

Next I am going to:

- Create a PR in `@types` to add the NodeJs Error properties like `code` in its `Error` interface.
- Create a new story to discuss 1. how to contribute a custom mapping in the reject action. 2. how to provide the mapping in a component. 3. update this example repo or the authentication module to add the mapping for the unauthorized error.

bajtos: > Create a PR in @types to add the NodeJs Error properties like code in its Error interface.

+1

Until that happens, you can use the following workaround:

```ts
const err = new HttpErrors.UnprocessableEntity('invalid email');
Object.assign(err, {code: 'INVALID_EMAIL'});
throw err;
```

> Create a new story to discuss 1. how to contribute a custom mapping in the reject action. 2. how to provide the mapping in a component. 3. update this example repo or the authentication module to add the mapping for the unauthorized error.

Awesome 👍

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+import {Request, HttpErrors} from '@loopback/rest';
+import {UserProfile} from '@loopback/authentication';
+import * as _ from 'lodash';
+import {AuthenticationStrategy} from './authentication.strategy';
+
+export class JWTStrategy implements AuthenticationStrategy {
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    // there is a discussion regarding how to retrieve the token,
+    // see comment https://github.com/strongloop/loopback-next/issues/1997#issuecomment-451054806
+    const token = request.query.token || request.headers['authorization'];
```
bajtos: In the real implementation, we should support Bearer scheme instead (or in addition) to sending the JWT key directly in `Authorization` header.

```
Authorization: Bearer header.payload.signature
```

See also https://security.stackexchange.com/questions/108662/why-is-bearer-required-before-the-token-in-authorization-header-in-a-http-re

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (token) {
```
bajtos: Please use reverse logic and early return to keep the code easier to read and less indented.

```ts
if (!token) {
  throw new HttpErrors.Unauthorized('Token not found!');
}
```

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+      try {
+        const decoded = await verifyAsync(token, SECRET);
+        let user = _.pick(decoded, ['id', 'email', 'firstName']);
+        (user as UserProfile).name = user.firstName;
+        delete user.firstName;
+        return user;
+      } catch (err) {
+        throw new HttpErrors.Unauthorized('Could not decode the JWT token!');
```
bajtos: This is discarding information about the original error: the reason why the token could not be decoded, the stack trace pointing to the problematic place.

I am proposing to use the following solution:

```ts
} catch (err) {
  Object.assign(err, {
    code: 'INVALID_ACCESS_TOKEN',
    statusCode: 401,
  });
  throw err;
}
```

src/controllers/user.controller.ts
```diff
@@ -11,15 +11,34 @@ import {hash} from 'bcryptjs';
...
+  authenticate,
+  UserProfile,
+  AuthenticationBindings,
+} from '@loopback/authentication';
+import {Credentials} from '../repositories/user.repository';
 
 const hashAsync = promisify(hash);
 
+// TODO(jannyHou): This should be moved to @loopback/authentication
+const UserProfileSchema = {
```
bajtos: We need to find a way allowing applications to add additional properties to user profile.

I would also prefer to leverage `@model` and `@property` decorators together with the existing code converting LB4 models to JSON/OpenAPI schema, instead of maintaining the schema manually.

```ts
@model
class UserProfile extends Model {
  @property({type: 'string', required: true})
  id: string;

  @property()
  email: string;

  @property()
  name: string;
}
```

Feel free to leave such changes out of scope of this spike and create a new user story to follow up on this area.

jannyHou: There will be a refactor PR in `@loopback/authentication` coming soon, this change will definitely be part of it :)

src/authentication-strategies/JWT.strategy.ts
jannyHou: Rewrote the test and function accordingly.

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,34 @@
...
+// Consider turn it to a binding
+const SECRET = 'secretforjwt';
+import {Request, HttpErrors} from '@loopback/rest';
+import {UserProfile} from '@loopback/authentication';
+import * as _ from 'lodash';
+import {AuthenticationStrategy} from './authentication.strategy';
+
+export class JWTStrategy implements AuthenticationStrategy {
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    let token = request.query.token || request.headers['authorization'];
```
raymondfeng: The query param is usually named as `access_token` instead of `token`.

src/providers/strategy.resolver.provider.ts
```diff
@@ -0,0 +1,28 @@
...
+// import {JWTStrategy} from '@loopback/authentication';
+export class StrategyResolverProvider
+  implements Provider<JWTStrategy | undefined> {
+  constructor(
+    @inject(AuthenticationBindings.METADATA)
+    private metadata: AuthenticationMetadata,
+  ) {}
+  value(): ValueOrPromise<JWTStrategy | undefined> {
+    if (!this.metadata) {
+      return Promise.resolve(undefined);
```
raymondfeng: Why not simply return `undefined`?

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!token) throw new HttpErrors.Unauthorized('Token not found!');
```
raymondfeng: `'No access token found'`?

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,35 @@
...
+// Consider turn it to a binding
+const SECRET = 'secretforjwt';
+import {Request, HttpErrors} from '@loopback/rest';
+import {UserProfile} from '@loopback/authentication';
+import * as _ from 'lodash';
+import {AuthenticationStrategy} from './authentication.strategy';
+
+export class JWTStrategy implements AuthenticationStrategy {
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    let access_token = request.query.token || request.headers['authorization'];
```
raymondfeng: @jannyHou I meant to say we should check `request.query.access_token`. The variable name can stay as `token` or `accessToken`.

src/controllers/user.controller.ts
```diff
@@ -87,4 +130,31 @@ export class UserController {
...
+          },
+        },
+      },
+    },
+  })
+  async login(
+    @requestBody() credentials: Credentials,
+  ): Promise<{token: string}> {
+    let res = {token: ''};
+    res.token = await this.userRepository.login(credentials);
```
raymondfeng: It's a bit awkward that the `login` method is implemented by `UserRepository`. Should we move it into the controller itself?

raymondfeng: The code can be simplified as:
```ts
const token = await this.userRepository.login(credentials);
return {
  token,
};
```

src/repositories/user.repository.ts
raymondfeng: This is a warning sign that `login` should be refactored out of the repo.

src/authentication-strategies/JWT.strategy.ts
jannyHou: got it.

src/controllers/user.controller.ts
jannyHou: @raymondfeng There are different opinions regarding it, see https://github.com/strongloop/loopback4-example-shopping/pull/26#discussion_r246745742
At first it was implemented in the controller, but then I hesitate, the good thing of having that in the repository is we don't make http calls to get the access token when test.

src/controllers/user.controller.ts
```diff
@@ -65,6 +88,30 @@ export class UserController {
...
+  async printCurrentUser(
+    @inject('authentication.currentUser') currentUser: UserProfile,
+  ): Promise<UserProfile> {
+    return Promise.resolve(currentUser);
+  }
+
+  // TODO(@jannyHou): missing logout function.
+  // as a stateless authentication method, JWT doesn't actually
+  // have a logout operation. See article for details:
+  // https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6
```
raymondfeng: We can still log out by invalidating the jwt token.

src/controllers/user.controller.ts
bajtos: Uh oh, was the implementation of login moved back to the controller? I consider that as anti-pattern, it breaks separation of concerns in MVC model:
 - Models/Repositories are responsible for data manipulation and domain model
 - Controllers are responsible for REST-specific behavior
 - Views are rendering the HTML output. In our case, we are building REST/JSON APIs, the view function is `JSON.stringify`.

Regarding login:
 - The repository layer should be responsible for preforming the actual login. There may be other scenarios besides REST where we want to login: integration tests, gRPC transport, websocket transport, etc.
 - The (REST) controller is responsible for obtaining login parameters from the request, calling the repository method to perform the actual login, and then converting the result into an HTTP response.

cc @raymondfeng

test/acceptance/user.controller.acceptance.ts
bajtos: I still see that test are performing login via REST API, not by calling Repository API.

@jannyHou PTAL again

jannyHou:
@bajtos 

> I still see that test are performing login via REST API, not by calling Repository API.

Yeah that's the refactor after moving the login function back to the controller.

@raymondfeng @bajtos Let's have a discussion about where to put the login logic:

I tend to agree with Raymond, my understanding is repository is only responsible for retrieving the data from backend then forming the domain object accordingly, things like generating the token should happen in the controller function...

cc @strongloop/loopback-next Any thought?

bajtos:
> I tend to agree with Raymond(B), my understanding is repository is only responsible for retrieving the data from backend then forming the domain object accordingly, things like generating the token should happen in the controller function...

Let's consider this use case: we want to implement `login` for REST clients, gRPC clients and WebSocket clients. Each transport uses different concepts, therefore we need one controller class for each transport.

When `login` is implemented fully inside the controller, then we need three mostly duplicated implementations of `login`, one for each controller. When `login` is implemented inside a repository class, each controller deals with transport-specific aspects only and the business logic of logging in is implemented in one shared place inside the Repository.

Another important aspect is testability. When `login` is implemented in a controller method, it's difficult to test in isolation. For example, to verify that incorrect credentials are rejected, we have to write an acceptance-level test going through all layers (REST Controller, Repository, Database). When `login` is implemented in a repository, we can write unit tests to verify how the REST Controller is exposing Repository's `login` method via REST and integration-level tests to verify that `login` handles all aspects of credential verification correctly.

> repository is only responsible for retrieving the data from backend then forming the domain object accordingly

IMO, repositories should be responsible for all business logic. Controllers should deal with transport-specific aspects only.

I see your point about repositories taking care of retrieving data objects from the backend only, maybe you are right that `login` does not belong to a Repository.

At the same time, I am strongly convinced that `login` implementation does not belong to a controller class either.

Maybe we are missing a third class? This class can receive `UserRepository` as a dependency, and provide methods like `login`, `logout`, etc.

bajtos:
> From testing perspective, it can be tested via both REST (integration) or the controller directly (unit).

I find it cumbersome that in order to verify that invalid email+password combination is rejected by `login`, the test has to deal with transport-specific details like how to send email+password credentials. I would like to see these two aspects decoupled:
 - Transport specific aspects go to controller method. This part is often unique for each transport.
 - The business logic of verifying credentials & issuing a new token should be implemented by a component that can be shared by multiple controllers and tested in isolation.

jannyHou:
> Let's consider this use case: we want to implement login for REST clients, gRPC clients and WebSocket clients. Each transport uses different concepts, therefore we need one controller class for each transport.
When login is implemented fully inside the controller, then we need three mostly duplicated implementations of login, one for each controller. When login is implemented inside a repository class, each controller deals with transport-specific aspects only and the business logic of logging in is implemented in one shared place inside the Repository.

This convinces me that the logic "generate the access token" should stay in a layer that could be shared among different clients. How about define it in service? Like the recommender service we have in https://github.com/strongloop/loopback4-example-shopping/tree/master/src/services.

Then each part's responsibility should be:

- Transport specific aspects like validating email+password should stay in the controller.
- The login logic -- takes in credentials, verify the user, generate the accesstoken, we extract it into a service which could be shared among different clients.
- The controller function calls the service to perform login.

Thought?

raymondfeng:
For login, there are a few parts involved. Here is what I see.

1. Extract the credentials from the request (controller)
2. Verify/validate the credentials  (a utility that possibly interacts with repository)
3. Generate the JWT access token if 2 is successful (a utility that possibly interacts with repository) 
4. Respond to clients (controller)

Personally, I don't think `login` should be anchored on the repository class, which is responsible for database access only.

bajtos:
> For login, there are a few parts involved. Here is what I see.
> 
> - Extract the credentials from the request (controller)
> - Verify/validate the credentials (a utility that possibly interacts with repository)
> - Generate the JWT access token if 2 is successful (a utility that possibly interacts with repository)
> - Respond to clients (controller)

I totally agree 👍 

> Personally, I don't think login should be anchored on the repository class, which is responsible for database access only.

As I am thinking about this more, placing `login` into a standalone class/utility will make the implementation easier to test in isolation, because the `login` business logic is decoupled from `repository` data access. It allows us to write unit-tests where the login utility is created with a mocked repository class.

@raymondfeng are we in agreement then to use the following design based on @jannyHou's comment above?

1. Transport specific aspects like extracting email+password from the request should stay in the controller.
2. The login logic -- takes in credentials, verify the user, generate the accesstoken, we extract it into a service which could be shared among different clients.
3. The controller function calls the service to perform login. The service is injected via DI.
4. The login service receives UserRepository via DI.
5. User repository remains unchanged.

codejamninja:
passport-jwt works great for jwt auth in loopback-4

src/authentication-strategies/JWT.strategy.ts
```diff
@@ -0,0 +1,34 @@
...
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    let token = request.query.access_token || request.headers['authorization'];
+    if (!token) throw new HttpErrors.Unauthorized('No access token found!');
+
+    if (token.startsWith('Bearer ')) {
+      token = token.slice(7, token.length);
+    }
+
+    try {
+      const decoded = await verifyAsync(token, SECRET);
```
bajtos: Based on our discussion around `login`, I feel this place should be changed in a similar way too.

- JWTStrategy deals with REST-specific aspects only: how to extract the token from the request, how to bind UserProfile as request's current user.
- Validation of the token and conversion into UserProfile should be handled by a new utility/service. This can be possibly the same class/file where `login` is implemented.

I expect the same benefits:
- Other transports like gRPC and WebSocket can use the same utility/service to validate access tokens and convert them into user profiles.
- The utility/service can be tested in isolation with no transport involved.

jannyHou:
@codejamninja Thank you for joining the discussion here, yeah I am trying to replace our implemented strategy with the community one you mentioned, then write a tutorial about how to plugin passport strategies in general, so that community could contribute their extensions/components to plugin custom strategies.

src/authentication-strategies/JWT.strategy.ts
jannyHou: @bajtos I removed the strategy written by us, replaced with the Strategy in `passport-jwt`, which is one of tens passport strategies that works with `passport` module out-of-box.

jannyHou:
@raymondfeng @bajtos I removed the strategy written by us, replaced with the Strategy in passport-jwt, which is one of tens passport strategies that work with passport module out-of-the-box. 

The new code is simpler than before. We only write a `StrategyResolverProvider`, bind it to `AuthenticationBindings.STRATEGY`, the custom `AuthenticateActionProvider` is not needed anymore.

Here a list of what to do/improve next:

- Make registering different strategies in `StrategyResolverProvider` extensible. 
- Extract the `UserProfile` related schemas/models into `@loopback/authentication`
- Add the logout function by invalidating jwt(I did a quick search while couldn't find a good way other than simply remove the token from the client...)
-  Write a tutorial of how to register a custom passport strategy and authorize endpoints with it.

jannyHou:
> Our codebase contains two JWT paths now. When creating a new token in our login function, we invoke jwt module directly. When authenticating a user, we invoke jwt via passport adapter (assuming the same underlying JWT implementation is used by us and passport-jwt, I did not check). 


@bajtos IIUC, your concern is the module `jsonwebtoken` and `passport-jwt` get out of sync later in the future? Module `passport-jwt` doesn't provide the `sign` function, that's why we have to invoke `jwt` directly when generate the access token.

Your concern is valid while I don't have a perfect solution to sync those two dependencies. And I think that should be a concern for all the `passport-jwt` users too.

> IMO, the AuthenticateActionProvider implemented by @loopback/authentication should be flexible enough to support both passport-based and custom strategies. Same applies to AuthenticationStrategy, it should be possible to use the one provided by @loopback/authentication together with our custom JWT implementation.

I totally agree...while I would suggest do it in a separate story. In my previous comment I suggested "Make registering different strategies in StrategyResolverProvider extensible.", I added one more for `AuthenticateActionProvider`.


I like all the discussions while to be honest since we start with implementing the jwt authentication in this example repository, and in this PR we gathered enough information about what to do next, I would like to land this PR soon(I understand it's not perfect but endless discussion wouldn't help move forward either IMO ☹️ ), we committed https://github.com/strongloop/loopback-next/issues/1997 and have https://github.com/strongloop/loopback-next/issues/1998 as stretch goal in this milestone, I would like to finish them as a priority. 

@bajtos I prefer to leverage `passport-jwt` because

- Give community a path to provide other passport based strategies.
- Sync `passport-jwt` and `jwt` would be a potential problem for all the `passport-jwt` users not only us.

While if you are strongly worried about "two implementations get out of sync" in the future, I am ok to revert the last commit. We can go back to the original implementation.

bajtos:
> I like all the discussions while to be honest since we start with implementing the jwt authentication in this example repository, and in this PR we gathered enough information about what to do next, I would like to land this PR soon(I understand it's not perfect but endless discussion wouldn't help move forward either IMO ☹️ )

I understand your frustration.

Let's take a look at #1997 description, emphasis is mine:

>  The goal of this step is to show **a reference implementation** in our Shopping App and prepare ground for the actual auth-related work in the next iterations. 
> (...)
> Ideally, there should be a documentation and/or a blog-post and/or a reference implementation **to make it easier for LB4 users to implement similar functionality in their project**.

What I would like to see as the outcome of this pull request, is a reference implementation that we are happy to recommend to all LB4 users as the right way of dealing with authentication.

Personally, I would prefer to revert the commit switching to passport-jwt:
- I believe that using passport strategies with LoopBack 4 has been solved long time ago by `@loopback/authentication`, no need to show that again.
- I see more benefits in showing people how to implement an authentication strategy without Passport, because that's not documented yet and I am not sure if`@loopback/authentication` actually supports that in the current version?
- Plus the concerns I expressed about introducing more dependencies in my earlier comments.

> IIUC, your concern is the module jsonwebtoken and passport-jwt get out of sync later in the future? Module passport-jwt doesn't provide the sign function, that's why we have to invoke jwt directly when generate the access token.
>
> Your concern is valid while I don't have a perfect solution to sync those two dependencies. And I think that should be a concern for all the passport-jwt users too.

I assumed that both `passport-jwt` and our code implementing token creation (signing) eventually call the same underlying module (`jsonwebtoken` or `jwt`, I don't know), and my primary concern was about getting into a state where each path (token creation vs. token verification & parsing) end up calling a different version of JTW implementation.

Let's consider two situations.

**Situation 1**
- A new, more secure signing algorithm is introduced.
- Our `login` implementation uses this new algorithm to sign the token created.
- passport-jwt is using an older version of JWT implementation that does not support the new signing algorithm yet.

If our authentication layer was calling JWT implementation directly, bypassing any additional abstractions like passport-jwt, then this situation would never happen.

**Situation 2**

A security vulnerability was found, for example in `jsonwebtoken` (a dependency of passport-js), or maybe researchers found a way how to compromise a certain signing algorithm and thus the algorithm is no longer considered as secure. Let's say the fix of this vulnerability requires a semver-major release of `jsonwebtoken`. 

How long will it take until `passport-js` is updated to use the new version of `jsonwebtoken` that is not vulnerable? According to npmjs.org, the last version of passport-js was published 10 months ago. See also the discussion in https://github.com/themikenicholson/passport-jwt/issues/151, it looks like this project has a single maintainer only.

---

Anyhow, this is just me. What do other @strongloop/loopback-maintainers think, especially @raymondfeng? Are they happy with recommending the current version of this pull request as "The Right Way"?

src/utils/user.authentication.ts
```diff
@@ -0,0 +1,43 @@
...
+import {HttpErrors} from '@loopback/rest';
+const jwt = require('jsonwebtoken');
+const signAsync = promisify(jwt.sign);
+
+export async function getAccessTokenForUser(
+  userRepository: UserRepository,
+  credentials: Credentials,
+): Promise<string> {
+  const foundUser = await userRepository.findOne({
+    where: {email: credentials.email, password: credentials.password},
```
bajtos: Please see the discussion in https://github.com/strongloop/loopback4-example-shopping/pull/26#discussion_r246744678 where I am explaining why storing passwords in plain-text is a security vulnerability.

To me, this is a blocker preventing landing of this pull request.

test/acceptance/user.controller.acceptance.ts
bajtos: This comment does not seem to be addressed, PTAL.

(A detail not preventing this PR from being approved.)

src/utils/user.authentication.ts
bajtos: On the second thought, I see that this problem was not introduced by your pull request, it was already there on `master`, so maybe it's alright to leave the fix out of scope of this pull request.

I am proposing the following approach:
- Create a new issue in this repository to keep track of this problem and allow us to plan the work in one of the next milestones.
- Here in this pull request, add a prominent code comment to point out the problem and refer people to the GitHub issue for further discussion.

test/acceptance/user.controller.acceptance.ts
jannyHou: @bajtos Hmm we already switched to `toJSON`, could you elaborate more about your expected implementation? thanks.

jannyHou:
@bajtos I reverted to the original implementation, let's focus on one solution in this PR :)

Some background of why I switched to `passport-jwt`: I saw an issue in loopback-next asking about how to implement another passport strategy and decorate endpoints with it, gave a quick research I found a  community module compatible with `passport`, but we don't have a documentation of how to plugin such modules with our existing `@loopback/authentication`. 
Considering that the community module `passport-jwt` is maturer than our implementation and people needs guide for plugin different passport strategies, I changed the implementation. 
Even though this PR doesn't leverage passport based strategies, I still suggest we add some tutorial/documentation about how to plugin them, maybe do it as part of "Make registering different strategies in StrategyResolverProvider extensible."

src/utils/user.authentication.ts
jannyHou: @bajtos Given a look of https://github.com/strongloop/loopback-next/issues/1996 I take back my word :( Let's address it in a separate story.

raymondfeng:
> @raymondfeng are we in agreement then to use the following design based on @jannyHou's comment above?

Yes. In addition to controller and repository, there should be a few classes (services - probably need to find a good name here) that can be bound to `Context` and use `DI`.

hacksparrow:
Do we want to include the `// Copyright IBM Corp. 2017,2018. All Rights Reserved. ...` comment on all the new files?

src/controllers/user.controller.ts
hacksparrow: @jannyHou we should have an endpoint which would invalidate the token for the user, technically "logging them out".

jannyHou:
@hacksparrow Good catch, we should

src/controllers/user.controller.ts
jannyHou: @hacksparrow I didn't find any good approach to invalidate the token other than removing it from client side, which is covered in test case https://github.com/strongloop/loopback4-example-shopping/pull/26/files#diff-cd803ccaed549218efa0937d5edd4533R179.
Any suggestions?

src/authentication-strategies/JWT.strategy.ts
bajtos: > Based on our discussion around `login`, I feel this place should be changed in a similar way too.
>
> - JWTStrategy deals with REST-specific aspects only: how to extract the token from the request, how to bind UserProfile as request's current user.
> - Validation of the token and conversion into UserProfile should be handled by a new utility/service. This can be possibly the same class/file where `login` is implemented.

@jannyHou I feel this comment should be addressed before landing the patch. If it makes it easier for you, then I am ok to defer this work to a follow-up pull request, as long as the change is made as part of the actual user story this pull request belongs to.

src/utils/user.authentication.ts
bajtos: Actually, the scope of strongloop/loopback-next#1996 is different. When I was creating that issue, I did not realize that we have a bigger problem in the current implementation!

As I see it, we have two changes to make:
- MUST HAVE: hash the passwords before they are stored in the database. I think we need to create a new issue for this?
- SHOULD HAVE: move credentials (password, salt?) from User model to a new model. This is covered by  strongloop/loopback-next#1996.

Please don't forget to add a comment to this place in code to warn users about not using this code as a reference implementation.

test/acceptance/user.controller.acceptance.ts
bajtos: Can we remove this line now? I believe that `toJSON` will take care of converting `id` from ObjectID to a string.

jannyHou:
The follow-up PRs and stories:

- [ ] (within the scope of #1997) Extract the utility/service functions in `JWT.strategy.ts` into utility files
- [ ] (within the scope of #1997) Write a tutorial/blog of how to add custom authentication strategy.
- [ ] (in progress) Refactor the UserProfile to be a model
- [ ] Hash the password (story to be created)
- [ ] Make registering different strategies in StrategyResolverProvider extensible. (story to be created)
- [ ] Make AuthenticateActionProvider flexible to support both passport based and other strategies. (story to be created)
- [ ] Add the logout function by invalidating jwt(I did a quick search while couldn't find a good way other than simply remove the token from the client...)

PR #32, opened 2019-01-25
refactor: extract decode user to util

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/1997
previous PR: https://github.com/strongloop/loopback4-example-shopping/pull/26

Refactor the decoding user code into a util function, so that it could be shared among different controllers other than REST.

### DIFF
--- a/src/utils/user.authentication.ts
+++ b/src/utils/user.authentication.ts
@@ -9,8 +9,10 @@ import {toJSON} from '@loopback/testlab';
 import {promisify} from 'util';
 import * as isemail from 'isemail';
 import {HttpErrors} from '@loopback/rest';
+import {UserProfile} from '@loopback/authentication';
 const jwt = require('jsonwebtoken');
 const signAsync = promisify(jwt.sign);
+const verifyAsync = promisify(jwt.verify);
 
 export async function getAccessTokenForUser(
   userRepository: UserRepository,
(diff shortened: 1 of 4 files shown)


src/utils/user.authentication.ts
```diff
@@ -46,3 +48,15 @@ export function validateCredentials(credentials: Credentials) {
     );
   }
 }
+
+// secret should be injected
```
jannyHou: This is implemented in the next PR https://github.com/strongloop/loopback4-example-shopping/pull/33

test/unit/utils.authentication.unit.ts
```diff
@@ -0,0 +1,65 @@
...
+    const expectedUser = getExpectedUser(newUser);
+    const currentUser = await decodeAccessToken(token, SECRET);
+    expect(currentUser).to.deepEqual(expectedUser);
+  });
+
+  it('throws error for invalid accesstoken', async () => {
+    const token = 'fake';
+    try {
+      await decodeAccessToken(token, SECRET);
+      expect('throws error').to.be.true();
```
bajtos: @jannyHou 

Here is a better way for verifying that a call of an async function throws an error:

```ts
await expect(decodeAccessToken(token, SECRET))
  .to.be.rejectedWith('jwt malformed');
```

jannyHou: @bajtos Thanks! I will address the feedback in https://github.com/strongloop/loopback4-example-shopping/pull/37

Issue #35, opened 2019-01-30
Store hashed password in the login function


jannyHou:
### Description

- [x] Create hashed password for authentication test cases like [this](https://github.com/strongloop/loopback4-example-shopping/blob/master/test/acceptance/user.controller.acceptance.ts#L141) merged as https://github.com/strongloop/loopback4-example-shopping/pull/37
- [x] Extract the hashing function into the utility file, or the service if PR https://github.com/strongloop/loopback4-example-shopping/pull/33 landed first

jannyHou:
Closed as https://github.com/strongloop/loopback4-example-shopping/pull/41 finished.

PR #37, opened 2019-01-30
fix: hash the password

jannyHou:
## Description 

Fix the accesstoken generation function by checking the password instead of sending the plain password.

A follow up PR for https://github.com/strongloop/loopback4-example-shopping/pull/26
And connect to the first acceptance criteria in https://github.com/strongloop/loopback4-example-shopping/issues/35

- [x] Local test passes

### DIFF
--- a/src/utils/user.authentication.ts
+++ b/src/utils/user.authentication.ts
@@ -10,6 +10,8 @@ import {promisify} from 'util';
 import * as isemail from 'isemail';
 import {HttpErrors} from '@loopback/rest';
 import {UserProfile} from '@loopback/authentication';
+import {compare} from 'bcryptjs';
+const compareAsync = promisify(compare);
 const jwt = require('jsonwebtoken');
 const signAsync = promisify(jwt.sign);
 const verifyAsync = promisify(jwt.verify);
@@ -19,10 +21,19 @@ export async function getAccessTokenForUser(
   credentials: Credentials,
 ): Promise<string> {
   const foundUser = await userRepository.findOne({
-    where: {email: credentials.email, password: credentials.password},
+    where: {email: credentials.email},
   });
   if (!foundUser) {
-    throw new HttpErrors.Unauthorized('Wrong credentials!');
+    throw new HttpErrors['NotFound'](
+      `User with email ${credentials.email} not found.`,
+    );
+  }
+  const passwordMatched = await compareAsync(
+    credentials.password,
+    foundUser.password,
+  );
+  if (!passwordMatched) {
+    throw new HttpErrors.Unauthorized('The credentials are not correct.');
   }
 
   const currentUser = _.pick(toJSON(foundUser), ['id', 'email', 'firstName']);
@@ -49,6 +60,8 @@ export function validateCredentials(credentials: Credentials) {
 }
 
 // secret should be injected
+// the refactor PR is in progress
+// https://github.com/strongloop/loopback4-example-shopping/pull/33
 export async function decodeAccessToken(
   token: string,
   secret: string,
--- a/test/acceptance/user.controller.acceptance.ts
+++ b/test/acceptance/user.controller.acceptance.ts
@@ -12,8 +12,11 @@ import {setupApplication} from './helper';
 import {createRecommendationServer} from '../../recommender';
 import {Server} from 'http';
 import * as _ from 'lodash';
+import {promisify} from 'util';
+import {hash} from 'bcryptjs';
 import {getAccessTokenForUser} from '../../src/utils/user.authentication';
 const recommendations = require('../../recommender/recommendations.json');
+const hashAsync = promisify(hash);
 
 describe('UserController', () => {
   let app: ShoppingApplication;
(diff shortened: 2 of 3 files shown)


src/utils/user.authentication.ts
```diff
@@ -18,11 +20,21 @@ export async function getAccessTokenForUser(
   userRepository: UserRepository,
   credentials: Credentials,
 ): Promise<string> {
+  console.log('getAccessTokenForUser', credentials.password);
```
raymondfeng: It should be removed.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -135,36 +138,40 @@ describe('UserController', () => {
...
 
   describe('authentication functions', () => {
-    // TODO: fix storing the plain password in the following issue:
-    // https://github.com/strongloop/loopback-next/issues/1996
+    let plainPassword: string;
+
+    before('create new user', async () => {
+      plainPassword = user.password;
+      // Salt + Hash Password
+      user.password = await hashAsync(user.password, 10);
```
raymondfeng: Should it be the responsibility of the controller/repository to hash passwords?

src/utils/user.authentication.ts
```diff
@@ -18,11 +20,21 @@ export async function getAccessTokenForUser(
...
   const foundUser = await userRepository.findOne({
-    where: {email: credentials.email, password: credentials.password},
+    where: {email: credentials.email},
   });
   if (!foundUser) {
-    throw new HttpErrors.Unauthorized('Wrong credentials!');
+    throw new HttpErrors.Unauthorized(
```
b-admike: Should we throw a 404 here or would that be a security risk?

src/utils/user.authentication.ts
```diff
@@ -18,11 +20,21 @@ export async function getAccessTokenForUser(
...
+      `User with email ${credentials.email} not found`,
+    );
+  }
+  const passwordMatched = await compareAsync(
+    credentials.password,
+    foundUser.password,
+  );
+  if (!passwordMatched) {
+    throw new HttpErrors.Unauthorized('The credential is not correct!');
```
b-admike: nitpick: I would rephrase this to `The credentials are not correct`.

test/acceptance/user.controller.acceptance.ts
b-admike: @jannyHou and I were talking about a hash utility function which can be called from the user repo or controller. I think it makes sense for the repository to call it when storing user passwords in the backend, but I see how that can be delegated to a controller too.

jannyHou: @raymondfeng @b-admike My opinion is unless we define a specific repository interface for model User and add a function for the hashing purpose, I tend to put the logic in controller. And as @b-admike said, we can extract it into utility/service that can be called from the controller.
Like what we have in [authentication utils](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/utils/user.authentication.ts).

jannyHou: And I plan to refactor them into the util in the next PR, this one focus on fixing the plain password.

Explained in the PR description :)

A follow up PR for #26
And connect to the first acceptance criteria in #35 (there will be another PR after it to refactor the hash utils into a proper place.)

src/utils/user.authentication.ts
jannyHou: @b-admike Good point, I think 404 makes more sense.

src/utils/user.authentication.ts
```diff
@@ -20,21 +20,21 @@ export async function getAccessTokenForUser(
   userRepository: UserRepository,
   credentials: Credentials,
 ): Promise<string> {
-  console.log('getAccessTokenForUser', credentials.password);
+  console.log(credentials.email);
```
b-admike: Can we remove the `console.log` altogether here?

jannyHou: good catch removed and squashed into one commit.

test/acceptance/user.controller.acceptance.ts
bajtos: @jannyHou Please do extract this code into utility/service that can be called from multiple places. For example tests.

test/unit/utils.authentication.unit.ts
```diff
@@ -40,15 +76,19 @@ describe('authentication', () => {
...
-      expect('throws error').to.be.true();
-    } catch (err) {
-      expect(err.message).to.equal('jwt malformed');
-    }
+    const error = new JsonWebTokenError('jwt malformed');
+    return expect(decodeAccessToken(token, SECRET)).to.be.rejectedWith(error);
   });
+
+  async function createUser() {
+    user.password = await hashAsync(user.password, 10);
```
bajtos: @jannyHou 
In the test suite, please use the smallest number that's still supported by bcrypt - we use `4` in LoopBack 3.x

The higher the number is, the longer it takes to compute the hash. Slow hash computation is good in production as it makes brute-force attacks difficult, but also impractical in tests because it can significantly slow down the test suite. We have been through this in LoopBack 3.x, let's not repeat that mistake again.

jannyHou: @bajtos Thanks for sharing the experience and explain the reason behind using the smallest number! 

>In the test suite, please use the smallest number that's still supported by bcrypt 

Will do it in the next PR that extracts hash functions into utils.

jannyHou: @bajtos fixed in https://github.com/strongloop/loopback4-example-shopping/pull/41/files#diff-40bd51c7c6fe0f144ef2295401976579R20

test/acceptance/user.controller.acceptance.ts
jannyHou: @bajtos See the refactor PR https://github.com/strongloop/loopback4-example-shopping/pull/41

Issue #39, opened 2019-02-01
Email and Password Validation 


psaunders:
This would be the perfect place to show people how to validate inputs using the methods that are hinted at in the documentation here: https://loopback.io/doc/en/lb4/Parsing-requests.html#request-body

> Please refer to the documentation on @requestBody decorator to get a comprehensive idea of defining custom validation rules for your models.

As the flagship example of the framework it doesn't look good having a bunch of old school if statements in the controller method doing validation. Someone who knows what they are doing could easily setup a "minimum length" example for the password which would help a lot.

jannyHou:
@psaunders 👍 Good catch.

jannyHou:
Acceptance Criteria:

- [ ]  Validate password by specifying `minLengh` property for `password` , see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties
- [ ] Validate email by specifying `format: email` for `email`, see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object-examples

psaunders:
Yeah you can put this in, but you end up with something pretty huge like this which is what we ended up with:

```ts
  @post('/accounts', {
    responses: {
      '200': {
        description: 'Account model instance',
        content: { 'application/json': { schema: { 'x-ts-type': Account } } },
      },
    },
  })
  async create(@requestBody({
      content: {
        'application/json': {
          schema: {
            required: ['email', 'password'],
            properties: {
              email: {
                type: 'string',
                format: 'email',
              },
              password: {
                type: 'string',
                pattern: '^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{5,512})',
                minLength: 8,
                maxLength: 512,
                not: {
                  enum: list,
                },
              },
            },
          },
        },
      },
    }) accountRequest: Account): Promise<Account> {
    if (!accountRequest.email || !accountRequest.password) { // Needed as the validator doesn't pass on the required nature of these
      throw new HttpErrors.UnprocessableEntity();
    }
    return this.accountService.createWithEmail(accountRequest.email!, accountRequest.password!); // Must be non-null due to validation
  }
```

It would be good if there was:
1. All the validation stuff added to the object so you could have this in your object class:
```
@property({
    type: 'string',
    minLength:8,
[...]

PR #41, opened 2019-02-06
refactor: extract hash password

jannyHou:
connect to https://github.com/strongloop/loopback4-example-shopping/issues/35

Refactor the password hash into utility functions.

### DIFF
--- a/src/controllers/user.controller.ts
+++ b/src/controllers/user.controller.ts
@@ -4,11 +4,9 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {repository} from '@loopback/repository';
-import {post, param, get, requestBody, HttpErrors} from '@loopback/rest';
+import {post, param, get, requestBody} from '@loopback/rest';
 import {User, Product} from '../models';
 import {UserRepository} from '../repositories';
-import {hash} from 'bcryptjs';
-import {promisify} from 'util';
 import {RecommenderService} from '../services/recommender.service';
 import {inject, Setter} from '@loopback/core';
 import {
@@ -20,10 +18,9 @@ import {Credentials} from '../repositories/user.repository';
 import {
   validateCredentials,
   getAccessTokenForUser,
+  hashPassword,
 } from '../utils/user.authentication';
-import * as isemail from 'isemail';
-
-const hashAsync = promisify(hash);
+import * as _ from 'lodash';
 
 // TODO(jannyHou): This should be moved to @loopback/authentication
 const UserProfileSchema = {
@@ -47,20 +44,8 @@ export class UserController {
 
   @post('/users')
   async create(@requestBody() user: User): Promise<User> {
-    // Validate Email
-    if (!isemail.validate(user.email)) {
-      throw new HttpErrors.UnprocessableEntity('invalid email');
-    }
-
-    // Validate Password Length
-    if (user.password.length < 8) {
-      throw new HttpErrors.UnprocessableEntity(
-        'password must be minimum 8 characters',
-      );
-    }
-
-    // Salt + Hash Password
-    user.password = await hashAsync(user.password, 10);
+    validateCredentials(_.pick(user, ['email', 'password']));
+    user.password = await hashPassword(user.password, 10);
 
     // Save & Return Result
     const savedUser = await this.userRepository.create(user);
--- a/src/utils/user.authentication.ts
+++ b/src/utils/user.authentication.ts
@@ -10,11 +10,18 @@ import {promisify} from 'util';
 import * as isemail from 'isemail';
 import {HttpErrors} from '@loopback/rest';
 import {UserProfile} from '@loopback/authentication';
-import {compare} from 'bcryptjs';
-const compareAsync = promisify(compare);
+import {genSalt, compare, hash} from 'bcryptjs';
 const jwt = require('jsonwebtoken');
 const signAsync = promisify(jwt.sign);
 const verifyAsync = promisify(jwt.verify);
+
+export async function hashPassword(
+  password: string,
+  rounds: number,
+): Promise<string> {
+  const salt = await genSalt(rounds);
+  return await hash(password, salt);
+}
 
 export async function getAccessTokenForUser(
   userRepository: UserRepository,
@@ -28,7 +35,7 @@ export async function getAccessTokenForUser(
       `User with email ${credentials.email} not found.`,
     );
   }
-  const passwordMatched = await compareAsync(
+  const passwordMatched = await compare(
     credentials.password,
     foundUser.password,
   );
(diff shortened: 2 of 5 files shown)


test/acceptance/user.controller.acceptance.ts
```diff
@@ -12,11 +12,11 @@ import {setupApplication} from './helper';
 import {createRecommendationServer} from '../../recommender';
 import {Server} from 'http';
 import * as _ from 'lodash';
-import {promisify} from 'util';
-import {hash} from 'bcryptjs';
-import {getAccessTokenForUser} from '../../src/utils/user.authentication';
+import {
+  getAccessTokenForUser,
+  hashPassword,
+} from '../../src/utils/user.authentication';
```
raymondfeng: We should try to avoid references to `src`. They should be imported from the module itself (assuming these functions are exported).

src/utils/user.authentication.ts
```diff
@@ -11,10 +11,19 @@ import * as isemail from 'isemail';
...
 const signAsync = promisify(jwt.sign);
 const verifyAsync = promisify(jwt.verify);
+const hashAsync = promisify(hash);
+
+export async function hashPassword(
+  password: string,
+  salt: number,
+): Promise<string> {
+  return await hashAsync(password, salt);
+}
```
raymondfeng: Out of scope for this PR: to be ideal, such utility functions should be bound to the context to support customization.

jannyHou:
FYI: The enhancement of validating credentials at the request level will be done in https://github.com/strongloop/loopback4-example-shopping/issues/39#issuecomment-461090909

test/acceptance/user.controller.acceptance.ts
jannyHou: @raymondfeng Could you explain more about what "module" means here? Is it the same thing as "artifact" like models, datasources...?

src/utils/user.authentication.ts
jannyHou: @raymondfeng Sounds reasonable to me, I have a PR for refactoring utilities into service:
https://github.com/strongloop/loopback4-example-shopping/pull/33
> utility functions should be bound to the context to support customization.

I think that PR is a good place to do ^

src/controllers/user.controller.ts
```diff
@@ -47,20 +44,8 @@ export class UserController {
...
-    if (user.password.length < 8) {
-      throw new HttpErrors.UnprocessableEntity(
-        'password must be minimum 8 characters',
-      );
-    }
-
-    // Salt + Hash Password
-    user.password = await hashAsync(user.password, 10);
+    validateCredentials(_.pick(user, ['email', 'password']));
+    user.password = await hashPassword(user.password, 4);
```
bajtos: - In production, we need to use many hash rounds to get high level of security. E.g. 10.
- In tests, we want to use as little rounds as allowed to get high speed. E.g. 4.

If it's too work to make this number customizable, then let's keep 10 for now and open a follow-up issue to address this properly.

jannyHou: @bajtos Oops I shouldn't change the round number in controller. Good catch. The util function takes in the round number, flexible enough to customize it. Reverted it to 10.

src/utils/user.authentication.ts
```diff
@@ -10,11 +10,20 @@ import {promisify} from 'util';
 import * as isemail from 'isemail';
 import {HttpErrors} from '@loopback/rest';
 import {UserProfile} from '@loopback/authentication';
-import {compare} from 'bcryptjs';
-const compareAsync = promisify(compare);
+import {compareSync, genSaltSync} from 'bcryptjs';
```
bajtos: Is it a good idea to use sync functions? I think this will block the event loop for non-negligible amount of time and can seriously degrade application's performance.

Can we use `compareAsync` instead? Same for `genSalt`.

jannyHou: @bajtos I got a problem when using the promisified `genSalt` with signature `genSaltAsync(10)`, opened a PR in `@types`, you can check the detail in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32868

That's the reason why I switched to `genSaltSync`, and I thought it will save the effort of doing promisify, therefore switched to sync styles for other functions.

> I think this will block the event loop for non-negligible amount of time and can seriously degrade application's performance.

But I now feel your point makes more sense. The time caused by blocking the event loop is more than the promisify time.

bajtos: > I got a problem when using the promisified `genSalt` with signature `genSaltAsync(10)`, opened a PR in `@types`, you can check the detail in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32868
>
> That's the reason why I switched to `genSaltSync`, and I thought it will save the effort of doing promisify, therefore switched to sync styles for other functions.

Until the typings are fixed, you can use an explicit type cast as a workaround. Just leave a comment explaining why the workaround is needed + a link to the GH issue you created.

For example:

```ts
const genSaltAsync = promisify(genSalt) as (rounds: number) => Promise<string>;
// or disable argument/retval type checks 
const genSaltAsync = promisify(genSalt) as Function;
```

We should get the implementation right and use async versions. If we need to bend the type system to do so, then let's do it.

src/utils/user.authentication.ts
```diff
@@ -10,11 +10,20 @@ import {promisify} from 'util';
...
 const signAsync = promisify(jwt.sign);
 const verifyAsync = promisify(jwt.verify);
+const hashAsync = promisify(hash);
+const compareAsync = promisify(compare);
+
+export async function hashPassword(
+  password: string,
+  rounds: number,
+): Promise<string> {
+  const salt = genSaltSync(rounds);
```
bajtos:
```suggestion
  const salt = await genSaltAsync(rounds);
```

test/acceptance/user.controller.acceptance.ts
raymondfeng: It's npm module. Something similar as `import {createRecommendationServer} from '../..';`, which points to the root of this app.

src/utils/user.authentication.ts
```diff
@@ -10,11 +10,20 @@ import {promisify} from 'util';
...
 import {HttpErrors} from '@loopback/rest';
 import {UserProfile} from '@loopback/authentication';
-import {compare} from 'bcryptjs';
-const compareAsync = promisify(compare);
+import {genSalt, compare, hash} from 'bcryptjs';
 const jwt = require('jsonwebtoken');
 const signAsync = promisify(jwt.sign);
 const verifyAsync = promisify(jwt.verify);
+const hashAsync = promisify(hash);
+const compareAsync = promisify(compare);
```
bajtos: I believe both `hash` and `compare` return promises when no callback was provided. Could you please try to `await` them directly without `promisify` wrapper?

API docs:

https://github.com/dcodeIO/bcrypt.js#hashs-salt-callback-progresscallback
> `@returns | Promise | If callback has been omitted`

https://github.com/dcodeIO/bcrypt.js#compares-hash-callback-progresscallback
> `@returns | Promise | If callback has been omitted`

Issue #40, opened 2019-02-01
Refactor authentication util functions into a service


jannyHou:
### Description

We currently keep authentication related util functions in a util file, refactoring them into a service will give us the benefit of dependency injection. 

> The controller function calls the service to perform login. The service is injected via DI.
The login service receives UserRepository via DI.

### Acceptance Criteria

- [x] Refactor the authentication related utils into a service class
- [x] The service is injected via DI in controller constructor, and controller function calls the service to perform login.
- [x] The UserRepository is injected via DI in service constructor.
- [x] Explorer better approach if the approach above is not the best practice

### PR in progress

https://github.com/strongloop/loopback4-example-shopping/pull/33

jannyHou:
Done.

PR #51, opened 2019-02-15
refactor: bcrypt hash

jannyHou:
A follow up PR for https://github.com/strongloop/loopback4-example-shopping/pull/33

connect to https://github.com/strongloop/loopback4-example-shopping/issues/40

### Description

- Refactor bcrypt hashing related functions into a service.
- Decouple the password hashing service from JWT service.

Details see discussion in https://github.com/strongloop/loopback4-example-shopping/pull/33#discussion_r255927845

### DIFF
--- a/src/controllers/user.controller.ts
+++ b/src/controllers/user.controller.ts
@@ -15,9 +15,9 @@ import {
   AuthenticationBindings,
 } from '@loopback/authentication';
 import {Credentials} from '../repositories/user.repository';
-import {HashPassword} from '../services/hash.password.bcryptjs';
+import {PasswordHasher} from '../services/hash.password.bcryptjs';
 import {JWTAuthenticationService} from '../services/JWT.authentication.service';
-import {JWTAuthenticationBindings, OtherServicesBindings} from '../keys';
+import {JWTAuthenticationBindings, PasswordHasherBindings} from '../keys';
 import {validateCredentials} from '../services/JWT.authentication.service';
 import * as _ from 'lodash';
 
@@ -39,16 +39,16 @@ export class UserController {
     public recommender: RecommenderService,
     @inject.setter(AuthenticationBindings.CURRENT_USER)
     public setCurrentUser: Setter<UserProfile>,
-    @inject(OtherServicesBindings.HASH_PASSWORD)
-    public hashPassword: HashPassword,
+    @inject(PasswordHasherBindings.PASSWORD_HASHER)
+    public passwordHahser: PasswordHasher,
     @inject(JWTAuthenticationBindings.SERVICE)
-    public jwt_authentication_service: JWTAuthenticationService,
+    public jwtAuthenticationService: JWTAuthenticationService,
   ) {}
 
   @post('/users')
   async create(@requestBody() user: User): Promise<User> {
     validateCredentials(_.pick(user, ['email', 'password']));
-    user.password = await this.hashPassword(user.password, 10);
+    user.password = await this.passwordHahser.hashPassword(user.password);
 
     // Save & Return Result
     const savedUser = await this.userRepository.create(user);
@@ -99,7 +99,6 @@ export class UserController {
   // as a stateless authentication method, JWT doesn't actually
   // have a logout operation. See article for details:
   // https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6
-
   @get('/users/{userId}/recommend', {
     responses: {
       '200': {
@@ -142,11 +141,12 @@ export class UserController {
       },
     },
   })
+  // @authenticate('jwt', {action: 'generateAccessToken'})
   async login(
     @requestBody() credentials: Credentials,
   ): Promise<{token: string}> {
     validateCredentials(credentials);
-    const token = await this.jwt_authentication_service.getAccessTokenForUser(
+    const token = await this.jwtAuthenticationService.getAccessTokenForUser(
       credentials,
     );
     return {token};
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -1,7 +1,7 @@
 import {BindingKey} from '@loopback/context';
 import {JWTAuthenticationService} from './services/JWT.authentication.service';
-import {HashPassword} from './services/hash.password.bcryptjs';
 import {JWTStrategy} from './authentication-strategies/JWT.strategy';
+import {PasswordHasher} from './services/hash.password.bcryptjs';
 
 // Discussion point for reviewers:
 // What would be the good naming conversion for bindings?
@@ -15,8 +15,9 @@ export namespace JWTAuthenticationBindings {
   );
 }
 
-export namespace OtherServicesBindings {
-  export const HASH_PASSWORD = BindingKey.create<HashPassword>(
-    'services.hash_password',
+export namespace PasswordHasherBindings {
+  export const PASSWORD_HASHER = BindingKey.create<PasswordHasher>(
+    'services.hasher',
   );
+  export const ROUNDS = BindingKey.create<number>('services.hasher.round');
 }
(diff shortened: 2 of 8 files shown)


src/keys.ts
```diff
@@ -15,8 +15,11 @@ export namespace JWTAuthenticationBindings {
   );
 }
 
-export namespace OtherServicesBindings {
-  export const HASH_PASSWORD = BindingKey.create<HashPassword>(
-    'services.hash_password',
+export namespace BcryptHasherBindings {
+  export const BCRYPT_HASHER = BindingKey.create<BcryptHasher>(
+    'services.bcrypt.hasher',
```
bajtos: In dependency inversion, the bindings are usually defined via interfaces, to allow consumers to provide any implementation matching the required contract.

I am proposing to change this binding to something like

```ts
export const PASSWORD_HASHER = BindingKey.create<PasswordHasher>(
  'services.password-hasher'
);
```

src/services/hash.password.bcryptjs.ts
```diff
@@ -17,3 +20,28 @@ export async function hashPassword(
   const salt = await genSalt(rounds);
   return await hash(password, salt);
 }
+
+export interface PasswordHasher<T = string> {
+  hashPassword(password: T): Promise<T>;
+  comparePassword(provided_pass: T, stored_pass: T): Promise<boolean>;
```
bajtos: We use camelCase, not snake_case.

```suggestion
  comparePassword(providedPass: T, storedPass: T): Promise<boolean>;
```
```

File: src/services/hash.password.bcryptjs.ts [hunk 3]
```diff
@@ -17,3 +20,28 @@ export async function hashPassword(
...
+  ) {}
+
+  async hashPassword(password: string): Promise<string> {
+    const salt = await genSalt(this.rounds);
+    return await hash(password, salt);
+  }
+
+  async comparePassword(
+    provided_pass: string,
+    stored_pass: string,
```
bajtos: Ditto.

src/keys.ts
```diff
@@ -15,8 +15,11 @@ export namespace JWTAuthenticationBindings {
...
+  );
+  export const ROUND = BindingKey.create<number>(
```
bajtos:
```suggestion
  export const ROUNDS = BindingKey.create<number>(
```

(plural "rounds", not a singular "round")

test/acceptance/user.controller.acceptance.ts
```diff
@@ -141,11 +141,17 @@ describe('UserController', () => {
   describe('authentication functions', () => {
     let plainPassword: string;
     let jwt_auth_service: JWTAuthenticationService;
+    let bcrypt_hasher: BcryptHasher;
 
     before('create new user', async () => {
+      bcrypt_hasher = new BcryptHasher(4);
```
bajtos: Since this is acceptance-level test, behavior should be configured mostly at app level via bindings. 

I think you should change `BcryptHasherBindings.ROUNDS` in one of the early hooks to ensure that all tests are using less rounds to hash password when creating new users.

test/acceptance/user.controller.acceptance.ts
```diff
@@ -141,11 +141,17 @@ describe('UserController', () => {
...
       plainPassword = user.password;
-      user.password = await hashPassword(user.password, 4);
-      jwt_auth_service = new JWTAuthenticationService(userRepo, JWT_SECRET);
+      user.password = await bcrypt_hasher.hashPassword(user.password);
```
bajtos: I am confused about this. Why is it necessary to hash the password here? Shouldn't UserRepository take care of hashing the password? At the moment, it's too easy to forget to hash the password before calling `userRepo.create` and end up with an invalid user record.

test/unit/utils.authentication.unit.ts
```diff
@@ -84,14 +85,19 @@ describe('authentication utilities', () => {
   });
 
   async function createUser() {
-    user.password = await hashPassword(user.password, 4);
+    bcrypt_hasher = new BcryptHasher(4);
+    user.password = await bcrypt_hasher.hashPassword(user.password);
```
bajtos: Similarly here. I would expect that the logic of password hashing should be implemented in a shared piece of code that's called both by runtime parts (repositories/controllers/services) and by tests.

Maybe I am missing something?

test/acceptance/user.controller.acceptance.ts
jannyHou: @bajtos The reason is the "hashing password" logic is handled in the user controller, NOT user repository.
When we create the user by making client request `POST /Users`, the controller will execute the hashing logic to make sure the hashed password is stored.
While when we call `userRepo.create(user)` directly, as what we did in the acceptance test, we need to hash the plain password first then store it.

test/unit/utils.authentication.unit.ts
jannyHou: @bajtos See https://github.com/strongloop/loopback4-example-shopping/pull/51#discussion_r258152498 ^

jannyHou:
@bajtos Thank you for the review.

> In dependency inversion, the bindings are usually defined via interfaces, to allow consumers to provide any implementation matching the required contract.

True I realized the implemented class is over used in most places, so I replaced it with interface `PasswordHasher` 

And regarding your concern about the test case, see my comment in https://github.com/strongloop/loopback4-example-shopping/pull/51#discussion_r258152498.

PTAL thanks.

src/services/hash.password.bcryptjs.ts
```diff
@@ -17,3 +20,28 @@ export async function hashPassword(
...
+    const salt = await genSalt(this.rounds);
+    return await hash(password, salt);
+  }
+
+  async comparePassword(
+    providedPass: string,
+    storedPass: string,
+  ): Promise<boolean> {
+    const passwordIsMatched = await compare(providedPass, storedPass);
+    return passwordIsMatched;
```
b-admike: nitpick: can we do `return await compare(providedPass, storedPass)` here?

src/controllers/user.controller.ts
```diff
@@ -39,16 +39,16 @@ export class UserController {
...
+    public passwordHahser: PasswordHasher,
     @inject(JWTAuthenticationBindings.SERVICE)
-    public jwt_authentication_service: JWTAuthenticationService,
+    public jwtAuthenticationService: JWTAuthenticationService,
   ) {}
 
   @post('/users')
   async create(@requestBody() user: User): Promise<User> {
     validateCredentials(_.pick(user, ['email', 'password']));
-    user.password = await this.hashPassword(user.password, 10);
```
b-admike: Do we want to set the rounds binding key to `10` for this change?

src/keys.ts
```diff
@@ -15,8 +15,9 @@ export namespace JWTAuthenticationBindings {
   );
 }
 
-export namespace OtherServicesBindings {
-  export const HASH_PASSWORD = BindingKey.create<HashPassword>(
-    'services.hash_password',
+export namespace PasswordHasherBindings {
+  export const PASSWORD_HASHER = BindingKey.create<PasswordHasher>(
+    'services.password.hasher',
```
b-admike: nitpick: Is it possible to rename `services.password.hasher` to `services.hasher.password` so that it's in the namespace `hasher` like `services.hasher.round` below?

src/controllers/user.controller.ts
jannyHou: @bajtos This `rounds` is set at the application level as 10 :) see https://github.com/strongloop/loopback4-example-shopping/pull/51/files#diff-8f42631f8d7d8cbaf47074d3ea5536d1R65

src/keys.ts
jannyHou: The difference is `services.password.hasher` maps to the hasher class, while `services.hasher.rounds` maps to a class member, maybe I should rename `services.password.hasher` to be `services.hasher`, does it make more sense?

b-admike: Your proposal sounds great, but feel free to ignore.

src/services/hash.password.bcryptjs.ts
jannyHou: @b-admike Sorry forgot to apply this change, I will remember to do it when refactor this example to use new authentication module.

PR #67, opened 2019-03-26
fix: add schema to describe credentials

jannyHou:
Screen Shot 2019-03-25 at 10 42 23 PM

Fix the input displayed in explorer of login function.


src/controllers/specs/user-controller.specs.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
```
nabdelgadir: Can you update the year?

jannyHou:
> Personally, I would use @model and @property decorators instead of writing the schema by hand

I can do it in story https://github.com/strongloop/loopback4-example-shopping/issues/39.
Yeah on a second thought I also tend to create a model...

Issue #43, opened 2019-02-08
user-order.controller.ts doesn't auth with @authenticate('jwt') 


lygstate:
user-order.controller.ts doesn't auth with @authenticate('jwt'),
besides, doesn't check the userId 
```
    if (userId !== currentUser.id) {
      throw new HttpErrors.BadRequest(
        `User id does not match: ${userId} !== ${currentUser.id}`,
      );
    }
```

jannyHou:
@lygstate 

```
if (userId !== currentUser.id) {
      throw new HttpErrors.BadRequest(
        `User id does not match: ${userId} !== ${currentUser.id}`,
      );
    }
```
This is an authorization scenario which we are planning to implement after building the fundamental extensible authentication system.

lygstate:
@jannyHou Sounds great, what's I can help.

jannyHou:
@lygstate Thank you for being interested to help!

The story that tracks this change is https://github.com/strongloop/loopback-next/issues/1998, we are planning to add some abstractions in `@loopback/authentication` module then refactor this example repo to leverage them. 
I will let you know if we decide story#1998 can start with the refactor in parallel or before it. And we can help you work on the PR that enable the auth for orders.

bajtos:
Before the work on the actual authorization is started, perhaps @lygstate can contribute the authentication changes we left out of February milestone? Cross-posting from https://github.com/strongloop/loopback-next/issues/1998:

> Rework `UserOrderController` to obtain the customer id from the request (the access token provided by the client) instead of a URL-path parameter.
> 
> For example, we can remove `/users/{userId}` prefix from all endpoints to end up with the following API:
>
> - `POST /orders` creates a new order
> - `GET /orders` returns all orders of the current user
> - `PATCH /orders?where=` to update some of the orders of the current user. I think this endpoint does not make sense in a Shopping app and should be eventually removed.
> - `DELETE /orders?where=` to delete some of the orders of the current user. I think this endpoint should be eventually removed, because orders are never deleted, they can be only closed (e.g. as cancelled).
>
> Existing REST API exposed by `UserController` should remain unauthenticated (allowing anonymous access).

jannyHou:
@bajtos Yeah sounds good 👍 
Hi @lygstate would you be interested to submit a PR according to the description cross posted in https://github.com/strongloop/loopback4-example-shopping/issues/43#issuecomment-468311127?

austin047:
[...]

PR #71, opened 2019-03-30
feat: addition of current user authentication in the request to creat…

austin047:
…e a new order

Updated the create order method that handles post request to create a user order to authenticate
against the current user before an order can be created, the client will have to pass the
authentication token in the request header which will be authenticated before the order is created

BREAKING CHANGE: For an order to be created the client must pass the authentication token in the
request header, otherwise a 401 Unauthorized error will be generated

feat #43

### DIFF
--- a/packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
+++ b/packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
@@ -89,6 +89,32 @@ describe('UserOrderController acceptance tests', () => {
       expect(res.body).to.deepEqual(order);
     });
 
+    it('creates an order for a user without a userId in the body', async () => {
+      const newUser = await userRepo.create(user);
+      const userId = newUser.id.toString();
+
+      const token = await jwtAuthService.getAccessTokenForUser({
+        email: newUser.email,
+        password: plainPassword,
+      });
+
+      const order = givenAOrder();
+
+      const res = await client
+        .post(`/users/${userId}/orders`)
+        .send(order)
+        .set('Authorization', 'Bearer ' + token)
+        .expect(200);
+      expect(res.body.orderId).to.be.a.String();
+      expect(res.body.userId).to.equal(userId);
+
+      delete res.body.orderId;
+      delete res.body.userId;
+      delete order.userId;
+
+      expect(res.body).to.deepEqual(order);
+    });
+
     it('throws an error when a userId in path does not match body', async () => {
       const newUser = await userRepo.create(user);
       const userId = newUser.id.toString();
--- a/packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
+++ b/packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
@@ -9,6 +9,11 @@ import {OrderRepository, UserRepository} from '../../repositories';
 import {MongoDataSource} from '../../datasources';
 import {User, Order} from '../../models';
 import {setupApplication} from './helper';
+import {JWTAuthenticationService} from '../../src/services/JWT.authentication.service';
+import {
+  PasswordHasherBindings,
+  JWTAuthenticationBindings,
+} from '../../src/keys';
 
 describe('UserOrderController acceptance tests', () => {
   let app: ShoppingApplication;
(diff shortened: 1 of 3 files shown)


src/controllers/user-order.controller.ts
```diff
@@ -35,10 +37,20 @@ export class UserOrderController {
...
   })
+  @authenticate('jwt')
   async createOrder(
     @param.path.string('userId') userId: string,
+    @inject('authentication.currentUser') currentUser: UserProfile,
     @requestBody() order: Order,
   ): Promise<Order> {
+    if (currentUser.id !== userId) {
+      throw new HttpErrors.BadRequest(
+        `User id does not match looged in user: ${userId} !== ${
```
emonddr: > User id does not match looged in user:

should be

User id does not match logged in user:

austin047: I have done the changes.

packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -35,10 +37,20 @@ export class UserOrderController {
       },
     },
   })
+  @authenticate('jwt')
   async createOrder(
     @param.path.string('userId') userId: string,
+    @inject('authentication.currentUser') currentUser: UserProfile,
     @requestBody() order: Order,
   ): Promise<Order> {
+    if (currentUser.id !== order.userId) {
```
jannyHou: The rewrite here is different from the expected behaviour:

The expected requirement mentioned [here](https://github.com/strongloop/loopback4-example-shopping/issues/43#issuecomment-468311127) is *don't take in `userId` from request*:

```ts
@authenticate('jwt')
async createOrder(
  @inject('authentication.currentUser') currentUser: UserProfile,
  @requestBody() order: Order,
) {
  if (!currentUser.id) {
    throw new HttpErrors.BadRequest(`No user logged in.`);
  }
  // The rest of code that creates orders for the logged in user.
}
```

But your logic which compares the logged in user id and the id from query also makes sense to me.

If aligning with the original requirements costs too much effort I am ok with the approach here.

Any thought @strongloop/loopback-next ?

packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
```diff
@@ -26,42 +31,113 @@ describe('UserOrderController acceptance tests', () => {
...
+    });
 
-    expect(res.body.orderId).to.be.a.String();
-    delete res.body.orderId;
-    expect(res.body).to.deepEqual(order);
-  });
+    it('creates an order for a user with a given orderId', async () => {
+      const newUser = await userRepo.create(user);
+      const userId = newUser.id.toString();
+      const order = givenAOrder({userId: userId, orderId: '1'});
```
jannyHou: nitpick: givenAnOrder

packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
```diff
@@ -26,42 +31,113 @@ describe('UserOrderController acceptance tests', () => {
...
+    it('creates an order for a user without a given orderId', async () => {
+      const newUser = await userRepo.create(user);
+      const userId = newUser.id.toString();
+
+      const token = await jwtAuthService.getAccessTokenForUser({
+        email: newUser.email,
+        password: plainPassword,
+      });
+
+      const order = givenAOrder({userId: userId});
```
jannyHou: any reason not provide the `orderId` here?

packages/shopping/src/controllers/user-order.controller.ts
samarpanB: +1. Agree. No user id in request.

bajtos: Yes please, we should really remove `userId` argument from this controller method.

In the implementation proposed by this pull request, there are too many places where user id can be specified:
 - `userId` as a path parameter
 - `order.userId` from request body
 - `currentUser.id` from the authentication layer

packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
bajtos: Ideally, I'd like the server to fill `userId` so that the clients making HTTP requests to create new orders don't have to. I.e. this test should pass with `const order = givenAnOrder()` too.

dhmlau:
@austin047, there are a few review comments from the maintainers.  Are you interested to continue to work on this PR? Thanks!

austin047:
Yes, I will work on it as soon as possible.

packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -35,13 +37,29 @@ export class UserOrderController {
       },
     },
   })
+  @authenticate('jwt')
   async createOrder(
     @param.path.string('userId') userId: string,
```
bajtos: Let's remove the `userId` argument please, we should take the id from `currentUser`.

dougal83:
Hey @austin047 Are you interested in continuing work on this PR?  I'll happily complete the work if you would prefer.  Thanks.

dougal83:
@jannyHou @hacksparrow should this PR be closed as inactive?

jannyHou:
@dougal83 Thank you for checking it, yes I think after refactoring the shopping app, all orders endpoints are secured in https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/controllers/user-order.controller.ts

To continue(or open a new PR), are you planning to remove the `/{userId}` from all paths? 

And really appreciate @austin047 's effort on updating it after auth was first added in this example. There are several breaking changes made since then.

dougal83:
@jannyHou Probably easier to close this PR and start afresh from the current master.  If you could create the story/issue with help wanted tag I'll pick it up when I have a moment unless someone beats me to it.  Thank you.

jannyHou:
@dougal83 Thank you created story in https://github.com/strongloop/loopback4-example-shopping/issues/629
Based on the chat I am going to close this PR. Let's keep discussion in #629. 
And apologize to @austin047 again for unable to merge his code 😞  we appreciate your effort spending on kicking off this refactor 💪

Issue #68, opened 2019-03-28
Move test files to `src/__tests__`


jannyHou:
## Description

Follow the new file structure discussed in https://github.com/strongloop/loopback-next/pull/2283, move the `/test` folder to `src/__tests__`.

Reference PR: https://github.com/strongloop/loopback-next/pull/2316

## Acceptance Criteria

- [ ] Refactor the test files into `src/__tests__`

dhmlau:
Fixed via https://github.com/strongloop/loopback4-example-shopping/pull/84

PR #82, opened 2019-04-16
refactor: move test files to src/__tests__

dhmlau:
Implements https://github.com/strongloop/loopback4-example-shopping/issues/68

Similar PR https://github.com/strongloop/loopback-next/pull/2316/files


package.json
```diff
@@ -28,10 +28,10 @@
     "tslint:fix": "npm run tslint -- --fix",
     "pretest": "npm run clean && npm run build && npm run docker:start",
     "pretest:ci": "npm run build",
-    "test": "lb-mocha --allow-console-logs \"dist/test\"",
-    "test:ci": "lb-mocha --allow-console-logs \"dist/test\"",
+    "test": "lb-mocha --allow-console-logs \"dist/src/__tests__/**/*.js\"",
```
dhmlau: In the similar PR https://github.com/strongloop/loopback-next/pull/2316, it references `dist/__tests__/**/*.js` without the `src` folder in the path, but somehow I cannot get it to work in here. 
cc @b-admike

nabdelgadir: So with your current changes, in the `dist` folder, it currently contains two folders: `src` and `recommender`. In the PR you referenced, in `tsconfig.build.json`, this change is made:
```diff
"compilerOptions": {
-   "rootDir": "."
+   "rootDir": "src"
},
```
Now all the files that were originally in the `src` folder are now the ones in the `dist` folder. The problem is that if you try to do that here, then it won't include the `recommender` folder.

dhmlau: right. that's the problem we're having now.  Tried below but still have the same error that "no test files found in dist/__tests__/**/*.js".
```
    "rootDirs": ["src", "recommender"]
```

raymondfeng: If we have more than one directories, they will be created under `dist`. We need to either make this repo a monorepo or refactor `recommender` into `src`.

dhmlau: Is the original intent of putting `recommender` folder outside `src` because it's not part of the LB application?  If our ultimate goal is to make this repo a monorepo where each "component" as separate microservices, maybe we'd go for the monorepo option?

dhmlau: After discussing it with @b-admike, I changed my mind.  Putting everything into `src` might be a simpler solution, and we can make this as the monorepo when we're going to make each component as microservices.

@strongloop/loopback-maintainers, thoughts?

bajtos: > Putting everything into src might be a simpler solution, and we can make this as the monorepo when we're going to make each component as microservices.

+1

raymondfeng:
See https://github.com/strongloop/loopback4-example-shopping/pull/84

dhmlau:
@raymondfeng , looks like ur PR #84 has already covered what I'm planning to do, so I'm going to abandon this PR. Thanks.

Issue #79, opened 2019-04-11
Refactor shopping app to use the updated authentication package


dhmlau:
## Description

1. Use the new authentication interface in `authentication` package => #2466 
2. Use the updated authentication action in `authentication` package => #2467 
3. Use the updated strategy resolver provider in `authentication` package => #2312 && #2467 (The resolver will need to deal with `passport` strategies and `non-passport`) strategies.
4. `/users/login` will continue to be without any `authenticate` decorator (won't have a strategy associated with it). email and password comes in the request body, and logic for performing user credential validation will continue to be done in the controller method.
5. The JWTStrategy will be reworked to be an extension of the authentication strategy extension point so that it is discoverable via the new strategy resolver provider in 3).
6. Take into account the community PR that added authorization to user orders.
     https://github.com/strongloop/loopback4-example-shopping/pull/71. Related to: https://github.com/strongloop/loopback-next/issues/1998

## Acceptance Criteria
- [ ] Updated documentation to guide user through steps of: starting database and redis, adding users to the database, logging in with email and password and receiving a jwt token, and using this jwt token with a REST Client (if API Explorer is not updated to pass credentials or tokens with requests) to access the authenticated endpoints, and using API Explorer for the non-authenticated endpoints
- [ ] Updated unit, acceptance, and integration mocha tests

JesusTheHun:
That would be so awesome if this was released this month ! ♥

bajtos:
@emonddr you may want to revive this pull request to upgrade example-shopping to v2 of our authentication module: https://github.com/strongloop/loopback4-example-shopping/pull/114 You can also run `git chery-pick 85098b6a` to add the commit from that pull request into your feature branch implementing the necessary changes and refactorings.

jannyHou:
🎉 PR merged.

PR #120, opened 2019-05-10
refactor: refactor to utilize latest @loopback/authentication module

emonddr:
Refactor this example shopping cart application to utilize the latest
authentication strategy interface, action provider, and
strategy provider from @loopback/authentication

associated with : https://github.com/strongloop/loopback4-example-shopping/issues/79


emonddr:
Waiting for code from https://github.com/strongloop/loopback-next/issues/2312 to land in a new release of @loopback/authentication 2.0 .

emonddr:
This PR is based on @loopback/authentication 2.0.3 , which was released yesterday by @b-admike  (thanks Biniam!)

packages/shopping/src/services/user-service.ts
```diff
@@ -0,0 +1,68 @@
...
+    // since first name and surname are optional, no error is thrown if not provided
+    let userName = '';
+    if (user.firstname && user.surname) {
+      userName = `${user.firstname} ${user.surname}`;
+    }
+
+    return {id: user.id, name: userName};
+  }
+
+  async validateCredentials(credentials: Credentials) {
```
b-admike: We don't have any async operations in this method, so we can remove the `async` keyword. You'd also have to change up the tests which expect it to behave in this manner.

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
```diff
@@ -29,63 +32,160 @@ describe('authentication services', () => {
...
+    const userProfile = userService.convertToUserProfile(newUser);
+    expect(expectedUserProfile).to.deepEqual(userProfile);
   });
 
-  it('decodeAccessToken throws error for invalid accesstoken', async () => {
-    const token = 'fake';
-    const error = new JsonWebTokenError('jwt malformed');
-    return expect(jwtService.decodeAccessToken(token)).to.be.rejectedWith(
-      error,
+  it('user service convertToUserProfile() succeeds without optional fields : firstname, surname', async () => {
```
b-admike: `convertToUserProfile` is not an async function, so let's remove the `async` keyword from the `it` block here.

package.json
```diff
@@ -48,8 +48,8 @@
     "@commitlint/config-conventional": "7.6.0",
     "@commitlint/config-lerna-scopes": "7.6.0",
     "@commitlint/travis-cli": "7.6.1",
-    "@loopback/build": "1.5.3",
-    "@loopback/testlab": "1.2.8",
+    "@loopback/build": "1.5.4",
+    "@loopback/testlab": "1.2.9",
```
raymondfeng: I wonder why the dependencies are declared with fixed versions instead of `^` range.

packages/shopping/src/__tests__/acceptance/shopping-cart.controller.acceptance.ts
```diff
@@ -18,11 +18,10 @@ describe('ShoppingCartController', () => {
   before('setupApplication', async () => {
     ({app, client} = await setupApplication());
   });
-
-  beforeEach(clearDatabase);
   after(async () => {
     await app.stop();
   });
+  beforeEach(clearDatabase);
```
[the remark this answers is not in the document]
raymondfeng: We can also argue that `before/after` (per suite) comes before `beforeEach/afterEach` (per test).

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -3,8 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Client, expect, toJSON} from '@loopback/testlab';
-import {Response} from 'superagent';
+import {Client, expect /*, toJSON*/} from '@loopback/testlab';
```
raymondfeng: If `toJSON` is not used, we should just remove it.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
-    after(async () => {
-      await recommendationService.stop();
+      const res = await client
+        .post('/users/login')
+        .send({email: user.email, password: user.password})
+        .expect(200);
+
+      const token = res.body.token;
+      expect(token).to.not.be.Null();
+      expect(token).to.be.String();
```
raymondfeng: This assertion covers `expect(token).to.not.be.Null();`

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
+      expect(token).to.not.be.empty();
+      const parts = token.split('.');
+      //token has 3 parts separated by '.'
+      expect(parts.length).to.equal(3);
```
raymondfeng: If we want to assert the format of the token, can we use a regular expression instead?

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
+        .expect(401);
+
+      expect(res.body.error.message).to.equal(
+        "Authorization header is not of type 'Bearer'.",
+      );
+    });
+
+    it('users/me returns an error when an expired token is provided', async () => {
+      const expiredToken =
+        'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjEiLCJuYW1lIjoiRXhhbXBsZSBVc2VyIiwiaWF0IjoxNTU3NTE1MTg1LCJleHAiOjE1NTc1MTUyNDV9.NZ31amMztvIYBth3SRY9fxiZTukObpgV2gSmJZ10pwY';
```
raymondfeng: Can we use a helper method to generate an expired token?

packages/shopping/src/application.ts
```diff
@@ -47,26 +45,23 @@ export class ShoppingApplication extends BootMixin(
...
+
+    this.bind(TokenServiceBindings.TOKEN_EXPIRES_IN).to('300');
+
+    this.bind(TokenServiceBindings.TOKEN_SERVICE).toClass(JWTService);
+
+    // // Bind bcrypt hash services
     this.bind(PasswordHasherBindings.ROUNDS).to(10);
     this.bind(PasswordHasherBindings.PASSWORD_HASHER).toClass(BcryptHasher);
 
+    this.bind(UserServiceBindings.USER_SERVICE).toClass(MyUserService);
```
raymondfeng: Maybe extract all bindings into a method?

packages/shopping/src/sequence.ts
```diff
@@ -33,12 +40,47 @@ export class MySequence implements SequenceHandler {
...
+      // resolve the strategy. When the strategy resolver obtains
+      // a strategy, it calls strategy.authentication(request) which
+      // is expected to return a user profile. If the user profile
+      // is undefined, then it throws a non-http error.
+      //
+      // It is necessary to catch these errors
+      // and rethrow them as http errors (in our REST application example)
+      //
+      // Errors thrown by the strategy implementations are http errors
+      // (in our REST application example). We simply rethrow them.
```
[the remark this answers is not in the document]
emonddr: @b-admike , I am referring to `packages/shopping/src/application.ts` .
My application is a REST application using http protocol, and so the errors that are thrown by my authentication strategy implementations are http errors. In the future, we can have a set of strategies that handle gRPC protocol requests, and the errors they would throw would be gRPC server errors, and the sequence would have to also catch the protocol-agnostic errors and rethrow them as gRPC errors.

packages/shopping/src/__tests__/acceptance/shopping-cart.controller.acceptance.ts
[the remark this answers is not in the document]
emonddr: @b-admike 
> Can we move this block back to where it was? Usually before or beforeEach blocks come before after and afterEach blocks 
Hi Biniam, I have the same opinion as @raymondfeng on this one.

package.json
jannyHou: I think it's because we don't have the shopping example as downstream test in loopback-next, so if new release breaks the example we won't know.

packages/shopping/src/__tests__/acceptance/shopping-cart.controller.acceptance.ts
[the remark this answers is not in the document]
b-admike: Oh okay I didn't see that, good point :-).

package.json
b-admike: It's because of how we configured renovate-bot (see https://github.com/strongloop/loopback4-example-shopping/pull/93#issuecomment-489586795 and #104).

packages/shopping/src/sequence.ts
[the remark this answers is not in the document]
b-admike: Oh okay, I'm mentioning it because I was a bit confused, so others might be too. If you can add what you've explained as a comment to users, that'd be great.

package.json
emonddr: @raymondfeng , @b-admike mentioned to me that `renovate-bot` goes into package.json and removes the `^` . So I just changed the versions to the latest.

emonddr: @b-admike did I paraphrase what you said properly?

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
-      await recommendationService.stop();
+      const res = await client
+        .post('/users/login')
+        .send({email: user.email, password: user.password})
+        .expect(200);
+
+      const token = res.body.token;
+      expect(token).to.not.be.Null();
+      expect(token).to.be.String();
+      expect(token).to.not.be.empty();
```
jannyHou: The token should always be a string I think, it's enforced by https://github.com/strongloop/loopback4-example-shopping/blob/4fdc4b2ee2bfdc7369a8c0b67c3d329ed5d555b2/packages/shopping/src/controllers/user.controller.ts#L136

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
+      const parts = token.split('.');
+      //token has 3 parts separated by '.'
+      expect(parts.length).to.equal(3);
     });
 
-    it('returns product recommendations for a user', async () => {
-      const newUser = await userRepo.create(user);
+    it('login returns an error when invalid email is used', async () => {
       await client
```
jannyHou: making http request is slower, please create the user using `const newUser = await userRepo.create(user);` to save time :)

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
emonddr: @raymondfeng , good catch. Just a silly comment I meant to leave temporarily while I was refactoring.

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
```diff
@@ -29,63 +32,160 @@ describe('authentication services', () => {
...
     );
+    expect(userProfile.id).to.equal(newUser.id);
+    expect(userProfile.name).to.equal('');
+  });
+
+  it('token service generateToken() succeeds', async () => {
+    const userProfile = userService.convertToUserProfile(newUser);
+    const token = await jwtService.generateToken(userProfile);
+    expect(token).not.Null();
+    expect(token).to.be.String();
```
jannyHou: Ditto https://github.com/strongloop/loopback4-example-shopping/pull/120/files#r284389151

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
```diff
@@ -29,63 +32,160 @@ describe('authentication services', () => {
...
+    const invalidToken = 'aaa.bbb.ccc';
+    await expect(jwtService.verifyToken(invalidToken)).to.be.rejectedWith(
+      expectedError,
+    );
+  });
+
+  it('password encrypter hashPassword() succeeds', async () => {
+    const encrypedPassword = await bcryptHasher.hashPassword(user.password);
+    expect(encrypedPassword).to.not.be.Null();
+    expect(encrypedPassword).to.be.String();
```
jannyHou: Ditto https://github.com/strongloop/loopback4-example-shopping/pull/120/files#r284389151

packages/shopping/src/authentication-strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,50 @@
...
+import {AuthenticationStrategy, UserProfile} from '@loopback/authentication';
+import {TokenServiceBindings} from '../keys';
+import {JWTService} from '../services/jwt-service';
+
+export class JWTAuthenticationStrategy implements AuthenticationStrategy {
+  name: string = 'jwt';
+
+  constructor(
+    @inject(TokenServiceBindings.TOKEN_SERVICE)
+    public tokenService: JWTService,
```
jannyHou: I think the type here should be [`TokenService`](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/token.service.ts#L11)

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -33,19 +38,27 @@ export class UserController {
     @inject.setter(AuthenticationBindings.CURRENT_USER)
     public setCurrentUser: Setter<UserProfile>,
     @inject(PasswordHasherBindings.PASSWORD_HASHER)
-    public passwordHahser: PasswordHasher,
-    @inject(JWTAuthenticationBindings.SERVICE)
-    public jwtAuthenticationService: JWTAuthenticationService,
+    public passwordHasher: PasswordHasher,
+    @inject(TokenServiceBindings.TOKEN_SERVICE)
+    public jwtService: JWTService,
```
jannyHou: Ditto, use type [`TokenService`](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/token.service.ts#L11)

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
emonddr: > If we want to assert the format of the token, can we use a regular expression instead?

As per our conversation offline, I will remove the test for 3 parts of the jwt token string, since this is tested later on when a jwt token is created and then used to go to /users/me endpoint.

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
```diff
@@ -29,63 +32,160 @@ describe('authentication services', () => {
...
-    const token = await jwtService.getAccessTokenForUser({
-      email: 'unittest@loopback.io',
-      password: 'p4ssw0rd',
-    });
-    expect(token).to.not.be.empty();
+  it('user service validateCredentials() succeeds', async () => {
+    const credentials = {email: 'dom@example.com', password: 'p4ssw0rd'};
+    return expect(
+      userService.validateCredentials(credentials),
+    ).to.be.fulfilled();
```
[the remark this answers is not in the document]
emonddr: @jannyHou I had to change this expect statement to the one you are mentioning to address the issue biniam raised. :)

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
emonddr: @jannyHou 
> making http request is slower, please create the user using const newUser = await userRepo.create(user); to save time :)

Hmm,  I don't agree. The purpose is to test the REST application and its endpoints, not test a repository.
The server needs to start anyways. I don't see the benefit or point of by-passing POST /users.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
       );
-      plainPassword = user.password;
-      user.password = await passwordHasher.hashPassword(user.password);
-      jwtAuthService = await app.get(JWTAuthenticationBindings.SERVICE);
     });
 
-    it('login returns a valid token', async () => {
-      const newUser = await userRepo.create(user);
+    it('login returns an error when invalid password is used', async () => {
       await client
```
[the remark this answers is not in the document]
emonddr: @jannyHou , please see my response .

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,162 @@ describe('UserController', () => {
...
     });
 
-    it('/me returns the current user', async () => {
-      const newUser = await userRepo.create(user);
-      const token = await jwtAuthService.getAccessTokenForUser({
-        email: newUser.email,
-        password: plainPassword,
-      });
+    it('users/me returns the current user profile when a valid token is provided', async () => {
+      let res = await client
```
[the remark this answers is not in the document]
emonddr: @jannyHou , please see my response .

packages/shopping/src/keys.ts
```diff
@@ -26,3 +25,11 @@ export namespace PasswordHasherBindings {
...
   export const ROUNDS = BindingKey.create<number>('services.hasher.round');
 }
+
+export namespace UserServiceBindings {
+  export const USER_SERVICE = BindingKey.create<MyUserService>(
+    'services.user.service',
+  );
+}
+
+import {PasswordHasher} from './services/hash.password.bcryptjs';
```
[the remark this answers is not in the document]
emonddr: @b-admike , why did me editor do that? weird. Good catch.

packages/shopping/src/sequence.ts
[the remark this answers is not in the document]
emonddr: @b-admike , this is explained in https://github.com/strongloop/loopback-next/blob/master/packages/authentication/docs/authentication-system.md , and in  in the @loopback/authentication module's mock sequence. I think the comments here explain it well enough.

@jannyHou and I plan to merge the  https://github.com/strongloop/loopback-next/blob/master/packages/authentication/docs files into the main [README.MD](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/README.md)

packages/shopping/src/authentication-strategies/jwt-strategy.ts
emonddr: @jannyHou , It can be either, but I will change it to the interface instead of the implementation.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
jannyHou: > The purpose is to test the REST application and its endpoints, not test a repository.

Ah true, but this test is for verifying endpoint `Users/login`, therefore it doesn't care about how the user is created, by calling repository API or sending HTTP request, by "making http request is slower" I mean how the user is created. 
Not a big deal though, I will leave it to you to decide how to create the user.

packages/shopping/src/authentication-strategies/jwt-strategy.ts
jannyHou: @emonddr As we discussed earlier, feel free to ignore the comment about changing the interface :)

package.json
b-admike: Yes @emonddr,  it'll remove them if we add the `^` back because of our settings.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
b-admike: I see @jannyHou's point. IIUC, we want to create the user using the less expensive repository operation, then we can use the our http client to make the request to `/users/login` which is what the test aims for. To see this in action, I'd run this test in isolation with both ways and see how much time it took (might have to change the spec to checkmarks instead of dots).

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,106 +96,155 @@ describe('UserController', () => {
...
+        .expect(401);
+
+      expect(res.body.error.message).to.equal(
+        'Error verifying token : jwt expired',
+      );
+    });
+  });
+
+  describe('user product recommendation (service) api', () => {
+    // tslint:disable-next-line:no-any
```
nabdelgadir: Is this line needed?

packages/shopping/src/authentication-strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,53 @@
...
+    public tokenService: TokenService,
+  ) {}
+
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    const token: string = this.extractCredentals(request);
+    const userProfile: UserProfile = await this.tokenService.verifyToken(token);
+    return userProfile;
+  }
+
+  extractCredentals(request: Request): string {
```
nabdelgadir: Nitpick: `extractCredentals` -> `extractCredentials`

packages/shopping/src/sequence.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2018. All Rights Reserved.
+// Copyright IBM Corp. 2019. All Rights Reserved.
```
nabdelgadir: Nitpick: `// Copyright IBM Corp. 2019. All Rights Reserved.` -> `// Copyright IBM Corp. 2018, 2019. All Rights Reserved.` (should be `{year created}, {year modified}`)

packages/shopping/src/sequence.ts
```diff
@@ -33,12 +40,47 @@ export class MySequence implements SequenceHandler {
     try {
       const {request, response} = context;
       const route = this.findRoute(request);
-      await this.authenticateRequest(request);
+
+      //
+      // The authentication action utilizes a strategy resolver to find
```
nabdelgadir: You can use a comment block instead but it's not a big deal so feel free to ignore

packages/shopping/src/services/user-service.ts
```diff
@@ -0,0 +1,52 @@
...
+      throw new HttpErrors.Unauthorized('The credentials are not correct.');
+    }
+
+    return foundUser;
+  }
+
+  convertToUserProfile(user: User): UserProfile {
+    // since first name and surname are optional, no error is thrown if not provided
+    let userName = '';
+    if (user.firstname && user.surname) {
```
nabdelgadir: Question: are they allowed to provide only `firstname` or `surname` or does it have to be both to be used as the `name`?

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
emonddr: @jannyHou , @b-admike explained to me the `specific tests` you wanted me to avoid using `POST /users`. I understand now. I thought you wanted me to avoid `POST /users` for every test.

packages/shopping/src/authentication-strategies/jwt-strategy.ts
emonddr: @jannyHou , I decided to go with the interfaces as you suggested.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
emonddr: @nabdelgadir , good catch. This is an old comment. This particular one is not needed, but there is another elsewhere that is needed. Good catch.

packages/shopping/src/services/user-service.ts
emonddr: @nabdelgadir . Just a personal decision at the time of writing. But you raise a good point.

raymondfeng: I would prefer `firstName` and `lastName`.

emonddr: @raymondfeng , I have addressed this in my latest commit

packages/shopping/src/__tests__/acceptance/helper.ts
```diff
@@ -27,3 +27,7 @@ export async function setupApplication(): Promise<AppWithClient> {
 
   return {app, client};
 }
+
+export function getExpiredToken(): string {
+  return 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjEiLCJuYW1lIjoiRXhhbXBsZSBVc2VyIiwiaWF0IjoxNTU3NTE1MTg1LCJleHAiOjE1NTc1MTUyNDV9.NZ31amMztvIYBth3SRY9fxiZTukObpgV2gSmJZ10pwY';
```
raymondfeng: Is there a way to use `jsonwebtoken ` api to create an expired token? If not, at least document how it was created at the first place with the corresponding json value.

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
```diff
@@ -26,66 +29,169 @@ describe('authentication services', () => {
...
+    const expectedUserProfile = {
+      id: newUser.id,
+      name: `${newUser.firstName} ${newUser.lastName}`,
+    };
+    const userProfile = userService.convertToUserProfile(newUser);
+    expect(expectedUserProfile).to.deepEqual(userProfile);
+  });
+
+  it('user service convertToUserProfile() succeeds without optional fields : firstName, lastName', () => {
+    const userWithoutFirstOrSurname = Object.assign({}, newUser);
```
raymondfeng: Use `lastName` for the var too.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
jannyHou: @b-admike Thank you for the explanation! 

> I thought you wanted me to avoid POST /users for every test.

Sorry I meant only those "specific tests", glad you guys figure it out, no misunderstanding now 👏

packages/shopping/src/__tests__/unit/utils.authentication.unit.ts
emonddr: Oops, nice catch! :)

packages/shopping/src/__tests__/acceptance/helper.ts
emonddr: @raymondfeng , using the utility, I created a token that expires in 5 minutes and printed out its value in console.log. Then, every other time I use it, it is always expired.

packages/shopping/src/services/jwt-service.ts is used to create the jwt token:
```
  async generateToken(userProfile: UserProfile): Promise<string> {
    if (!userProfile) {
      throw new HttpErrors.Unauthorized(
        'Error generating token : userProfile is null',
      );
    }

    // Generate a JSON Web Token
    let token: string;
    try {
      token = await signAsync(userProfile, this.jwtSecret, {
        expiresIn: Number(this.jwtExpiresIn),
      });
    } catch (error) {
      throw new HttpErrors.Unauthorized(`Error encoding token : ${error}`);
    }

    return token;
  }
```

emonddr:
squashed commits

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -100,110 +105,182 @@ describe('UserController', () => {
...
+    passwordHasher = await app.get(PasswordHasherBindings.PASSWORD_HASHER);
+  }
+
+  /**
+   * Creates an expired token
+   *
+   * Specifying a negative value for 'expiresIn' so the
+   * token is automatically expired
+   */
+  async function givenAnExpiredToken() {
```
raymondfeng: 👍

packages/shopping/src/__tests__/acceptance/helper.ts
emonddr: packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts now contains :
givenAnExpiredToken()

PR #136, opened 2019-05-27
fix(loopback4-example-shopping): add unique email constraint to user …

### LINKED ISSUE #103: Login issue: different user can use the same email.
I can create different users with the same email. Because users are using email for login, that could be a problem.

Example:
I create two users with the same email but different password:
```
{
    "email": "duplicate@gmail.com",
    "password": "test123456",
    "firstname": "t1",
    "surname": "t1"
}

{
    "email": "duplicate@gmail.com",
    "password": "test1234567",
    "firstname": "t2",
    "surname": "t2"
}
```

Then I can only login as the first user. If I try the second user's password, I get UnauthorizedError:

```
{
    "error": {
        "statusCode": 401,
        "name": "UnauthorizedError",
        "message": "The credentials are not correct."
    }
}
```
The second user can never login.

dougal83:
…model

add unique index to email in user model in file user.model.ts

fix #103

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "commitmsg": "commitlint -E GIT_PARAMS",
     "docker:start": "./bin/start-dbs.sh",
     "docker:stop": "./bin/stop-dbs.sh",
+    "migrate": "node ./packages/shopping/dist/migrate",
     "lint": "npm run prettier:check && npm run eslint",
     "lint:fix": "npm run prettier:fix && npm run eslint:fix",
     "prettier:cli": "lb-prettier \"**/*.ts\" \"**/*.js\" \"**/*.md\"",
--- a/packages/shopping/package.json
+++ b/packages/shopping/package.json
@@ -17,6 +17,7 @@
     "clean": "lb-clean dist*",
     "docker:start": "../../bin/start-dbs.sh",
     "docker:stop": "../../bin/stop-dbs.sh",
+    "migrate": "node ./dist/migrate",
     "pretest": "npm run clean && npm run build && npm run docker:start",
     "pretest:ci": "npm run build",
     "test": "lb-mocha --allow-console-logs \"dist/__tests__/**/*.js\"",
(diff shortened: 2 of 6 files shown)


bajtos:
Thank you for the pull request. Can you add a test please?

dougal83:
@bajtos Sorry, I will add tests and actually commit the controller additions.  My mistake.  Quick question before I resubmit.  What would be your implementation recommendation, should I rely on the database to enforce constraint(as I originally intended) or should I search for existing email in controller?

dougal83:
Since this is an example I opted to add check to the controller.  Seemed simpler than to involve the db migrate functionality.

bajtos:
We should rely on the database, that's the only reliable option.

The solution you are proposing at controller level has two problems:

(1) It's possible to write code that's calling `this.userRepository.create` (e.g. in a new controller method, or perhaps a script seeding database) and bypassing your check. At minimum, the check needs to be implemented in `UserRepository`.

(2) But even then, there is a race condition. If two requests to create a new user arrive at the same time, it's possible that they are handled in the following order:

1. Request1 arrives, sends `count` query to the database
2. Request2 arrives, sends `count` query to the database
3. `count` returns 0, Request1 proceeds to create a new user
4. `count` returns 0, Request2 proceeds to create a new user
5. BOOM. We have two users with the same email.

See also https://github.com/strongloop/loopback-next/issues/2331 where we are discussing the problem of referential integrity & unique constraints, and related items like https://github.com/strongloop/loopback-next/issues/2606 & https://github.com/strongloop/loopback-next/pull/2712.

bajtos:
IMO, the solution in  994ba09 was the right one.

IIRC, we are using MongoDB to store the data and it is able to define unique constraints. So let's  call `automigrate` from a `before` hook to ensure the database is properly set up for the tests.

dougal83:
@bajtos PTAL

I think I need to correct the commit message but is it OK to assume the MongoError 11000 duplicate key is email at this point?  Also is the test OK for double post or is there a better way to double post the same user?

Also thank you for collating all the relevant posts.  Very helpful.

dougal83:
Right I'll stop fiddling, lack of precision was irking me. Refactored to use a named index (if there is a better way please enlighten me): [3c76834](https://github.com/strongloop/loopback4-example-shopping/commit/f5f18be8c9bd08c49d7416b25f409a2493c76834) was the right one.

PR #231, opened 2019-08-08
feat: add authorization to user/order route

jannyHou:
issue: https://github.com/strongloop/loopback-next/issues/3695

implements loopback features in https://github.com/strongloop/loopback-next/pull/1205 and https://github.com/strongloop/loopback-next/issues/2900

The PR serves as a PoC and feature to verify implementation of [@loopback/authorization](https://github.com/strongloop/loopback-next/pull/1205) in the shopping app.

items:
- [X] - Try apply the casbin based authorization in the shopping example
- [X] - Figure out the responsibilities of `authentication`, `authorization`, `authorizeInterceptor`, `authorizor`, `fixtures(casbin)`
- [X] - Extract the common layer for `authentication` and `authorization`
- [X] - implement one endpoint `Users/{userId}/orders` with casbin policies , connect with loopback feature https://github.com/strongloop/loopback-next/pull/3529
- [x] - ~To be improved: casbin policies set fixed roles instead of checking the `order`'s ownId on the fly.~ Solved by adding a voter.
- [X] - add acceptance and unit tests


packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -41,10 +43,13 @@ export class UserOrderController {
...
     },
   })
+  @authenticate('jwt')
+  @authorize({resource: 'order', scopes: ['create']})
   async createOrder(
     @param.path.string('userId') userId: string,
     @requestBody() order: Order,
   ): Promise<Order> {
+    // should be moved to an authorizor
     if (userId !== order.userId) {
```
jannyHou: or use voter to decide.

packages/shopping/package.json
```diff
@@ -15,8 +15,8 @@
     "build:watch": "lb-tsc --watch",
     "clean": "lb-clean dist* *.tsbuildinfo",
     "migrate": "node ./dist/migrate",
-    "pretest": "npm run clean && npm run build",
     "pretest:ci": "npm run build",
+    "pretest": "npm run clean && npm run build && npm run docker:start",
```
raymondfeng: Please remove ` && npm run docker:start`

packages/shopping/package.json
```diff
@@ -40,17 +40,20 @@
...
+    "@loopback/core": "^1.10.6",
+    "@loopback/openapi-v3": "1.9.1",
+    "@loopback/repository": "1.10.1",
+    "@loopback/rest": "1.16.6",
+    "@loopback/rest-explorer": "1.3.1",
+    "@loopback/security": "^0.1.6",
+    "@loopback/service-proxy": "1.3.0",
+    "@types/jsonwebtoken": "8.3.2",
     "bcryptjs": "2.4.3",
+    "casbin": "^3.0.3",
```
raymondfeng: Please remove `^` from all versions.

packages/shopping/src/services/authorizor.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * @returns authenticateFn
+   */
+  value(): Authorizer {
+    return this.authorize.bind(this);
+  }
+
+  async authorize(
+    // @jannyHou: make sure it's universal for authorizers
```
dhmlau: is it a reminder for something?  Either way, I think it's the best to remove the `@jannyHou`. :)

packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
```diff
@@ -35,8 +42,11 @@ describe('UserOrderController acceptance tests', () => {
     const userId = user.id;
     const order = givenAOrder({userId: userId, orderId: '1'});
 
+    const token = await authenticateUser(user);
```
jannyHou: I think we can just generate the token once, then use it in all the `it` tests.

packages/shopping/src/services/enforcer.ts
```diff
@@ -2,7 +2,10 @@ import * as casbin from 'casbin';
...
 export async function createEnforcer() {
-  const conf = path.resolve(__dirname, '../fixtures/casbin/rbac_model.conf');
-  const policy = path.resolve(__dirname, '../fixtures/casbin/rbac_policy.csv');
-  return await casbin.newEnforcer(conf, policy);
+  const conf = path.resolve(__dirname, '../../fixtures/casbin/rbac_model.conf');
+  const policy = path.resolve(
+    __dirname,
+    '../../fixtures/casbin/rbac_policy.csv',
+  );
+  return casbin.newEnforcer(conf, policy);
```
jannyHou: good catch, and we can turn `createEnforcer` into a sync function since `casbin.newEnforcer()` is sync.

packages/shopping/src/services/jwt-service.ts
```diff
@@ -54,7 +58,7 @@ export class JWTService implements TokenService {
       );
     }
     const userInfoForToken = {
-      id: userProfile[securityId],
+      id: userProfile.id,
```
jannyHou: I think it should still be `userProfile[securityId]`

packages/shopping/src/services/authorizor.ts
jannyHou: Good catch...I think it's some notes from meeting, should be removed.

packages/shopping/package.json
deepakrkris: agreed

packages/shopping/src/__tests__/acceptance/user-order.controller.acceptance.ts
deepakrkris: actually every test iteration calls givenAUser() to create a new user for every test. so a token has to be created every time for the new user.

packages/shopping/src/services/authorizor.ts
deepakrkris: agreed, removing note

packages/shopping/src/services/enforcer.ts
deepakrkris: casbin.newEnforcer returns a promise, so to resolve it createEnforcer has to be async. The reason I removed the await is because of lint errors and await is added in a return from an async function internally.

packages/shopping/package.json
deepakrkris: agreed

packages/shopping/src/services/jwt-service.ts
deepakrkris: userProfile.id and userProfile[securityId] have same values.

deepakrkris: depends on what is returned by user service 
```
  convertToUserProfile(user: User): UserProfile {
    // since first name and lastName are optional, no error is thrown if not provided
    let userName = '';
    if (user.firstName) userName = `${user.firstName}`;
    if (user.lastName)
      userName = user.firstName
        ? `${userName} ${user.lastName}`
        : `${user.lastName}`;
    return {[securityId]: user.id, name: userName, id: user.id};
  }
```

packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -82,6 +91,8 @@ export class UserOrderController {
       },
     },
   })
+  @authenticate('jwt')
+  @authorize({resource: 'order', scopes: ['patch']})
```
jannyHou: ditto

packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -66,6 +73,8 @@ export class UserOrderController {
       },
     },
   })
+  @authenticate('jwt')
+  @authorize({resource: 'order', scopes: ['find']})
```
jannyHou: can we also add `voters: [compareId]` for patch and find methods?

packages/shopping/src/services/jwt-service.ts
jannyHou: true in this case theoretically `[securityId]` always equals to `id`, but conceptually we should retrieve the value from a user profile's unique field, which is `[securityId]`

deepakrkris: ok, I will correct this.

packages/shopping/src/controllers/user-order.controller.ts
deepakrkris: agreed

packages/shopping/src/controllers/user-order.controller.ts
deepakrkris: agreed

packages/shopping/package.json
```diff
@@ -40,17 +40,20 @@
...
+    "@loopback/authentication": "^3.2.1",
+    "@loopback/authorization": "^0.4.3",
+    "@loopback/boot": "^1.5.10",
+    "@loopback/context": "^1.23.4",
+    "@loopback/core": "^1.10.6",
+    "@loopback/openapi-v3": "1.9.1",
+    "@loopback/repository": "1.10.1",
+    "@loopback/rest": "1.16.6",
+    "@loopback/rest-explorer": "1.3.1",
+    "@loopback/security": "^0.1.6",
```
raymondfeng: Please remove all leading `^` for version ranges.

jannyHou: @deepakrkris ^ +1 and please also resolve the conflicts for 2 package-lock.json files. LGTM after cleaning up the package.json file :shipit:

deepakrkris: agreed

PR #237, opened 2019-08-11
refactor: correct misleading variable name

nflaig:
Correct naming from `decryptedToken` to `decodedToken`. The prior naming is misleading as it implies
that the token is encrypted instead of encoded.


dhmlau:
@nflaig, thanks again for another contribution! Your PR has landed. 🎉

PR #267, opened 2019-09-09
feat: enable swagger-ui token

jannyHou:
The spike PR for https://github.com/strongloop/loopback-next/issues/2027
Details are explained in the readme file.


hacksparrow:
With the limitations and the next plans explained in the docs, this is looking good. Excellent job @jannyHou 👍

README.md
```diff
@@ -110,13 +104,75 @@ possible by the use of the `UserService` service provided by
...
 
-### Tutorial
+## Authentication
+
+### Enabling Auth Token with Swagger-ui
+
+`swagger-ui` module is built with authorization component, which will show up by
+adding the security schema and operation security spec in the OpenAPI spec.
+
+You can check the swagger
```
bajtos:
```suggestion
You can check the OpenAPI
```

packages/shopping/src/index.ts
```diff
@@ -5,12 +5,18 @@
...
 export {ShoppingApplication, PackageInfo, PackageKey} from './application';
 
 export async function main(options?: ApplicationConfig) {
   const app = new ShoppingApplication(options);
 
   await app.boot();
+  let oaiSchema = app.getSync(RestBindings.API_SPEC);
+  addSecuritychema(oaiSchema);
+  console.log(oaiSchema.components!.securitySchemes);
+  app.bind(RestBindings.API_SPEC).to(oaiSchema);
```
bajtos: Can we use `app.api()` instead of this low-level implementation manipulating bindings?

See
https://github.com/strongloop/loopback-next/blob/70ad2ced8119ddf4b7da28ced5643edd4c5084de/packages/rest/src/rest.application.ts#L276-L289

Something along the following lines:

```ts
this.api({
  components: {
    ...BEARER_SECURITY_SCHEMA_SPEC,
  },
  security: {
    ...SECURITY_SPEC,
  }
});
```

Either way, the security schema should be configured from application constructor, not from the `main` function. It should be possible to start the application by calling `app.boot()` and `app.start()` directly (no `main` involved) and still have security schemas defined.

As a general rule, the `main` function should have a single responsibility - setup & start the application. Everything else should be handled by app's constructor or `boot`/`start`methods.

packages/shopping/src/application.ts
```diff
@@ -99,4 +100,17 @@ export class ShoppingApplication extends BootMixin(
...
+  async start(): Promise<void> {
+    await super.start();
+    const oaiSpec = this.getSync(RestBindings.API_SPEC);
+    this.api(
+      Object.assign(oaiSpec, {
+        components: {
+          ...SECURITY_SCHEMA_SPEC,
+        },
+        security: SECURITY_SPEC,
+      }),
```
bajtos: I believe it should be fine to call `this.api()` just with Security parts, because the code gathering spec from controllers & routes will merge that new spec with the object supplied to `this.api()`.

Have you tried to call `this.api()` from the constructor? 

```ts
export class ShoppingApplication extends BootMixin(
  ServiceMixin(RepositoryMixin(RestApplication)),
) {
  constructor(options?: ApplicationConfig) {
    super(options);

    this.api({
      components: {
          ...SECURITY_SCHEMA_SPEC,
      },
      security: SECURITY_SPEC,
  });

  // etc.
}
```

In my opinion, the security spec does not depend on the app being started, therefore it should be available even before `app.start()` is called. Thoughts?

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -83,6 +95,7 @@ export class UserController {
   }
 
   @get('/users/{userId}', {
+    security: SECURITY_SPEC_OPERATION,
```
bajtos: IIUC, this is configuring `basicAuth` schema. Is that intentional? I thought that we use `bearerAuth` schema everywhere. Or is this just for the spike?

packages/shopping/src/application.ts
bajtos: Should we perhaps add a story to improve `app.api()` to intelligently merge the new spec with any spec provided by previous calls of `app.api()`?

packages/shopping/src/controllers/user.controller.ts
jannyHou: @bajtos Oh this is just a demo, for spike only, I just want to verify that a operation level security policy(which is different than the global `bearerAuth`) works.

>I thought that we use bearerAuth schema everywhere

Exactly 💯

packages/shopping/src/application.ts
jannyHou: @bajtos Now `app.api()` takes in an entire spec in type `OpenAPISpec`, not `Partial<OpenAPISpec>`, that's why it's not possible to merge a partial spec into the generated server's spec.

Do we want to change its behaviour to take in a **partial** OpenAPI spec? If so I can create a story for it :)

README.md
```diff
@@ -110,13 +104,126 @@ possible by the use of the `UserService` service provided by
...
+button called "Authorize" on the right corner of the explorer:
+
+![authorize-button](/imgs/authorize-button.png)
+
+Click it and the token set dialog will be prompted:
+
+![set-token](/imgs/set-token.png)
+
+Paste the token you just copied in the field, then click "Authorize". The token
+will be be hidden:
```
agnes512: -> will be hidden

README.md
```diff
@@ -110,13 +104,126 @@ possible by the use of the `UserService` service provided by
...
+    responses: RESPONSE_SPEC,
+  })
+  async findById(@param.path.string('userId') userId: string): Promise<User> {
+    // Print out the header, you will see the Basic auth string in the header
+    console.log(`header ${inspect(this.request.headers)}`);
+    // business logic
+  }
+```
+
+The "Authorize" dialog will show up your new entry as:
```
agnes512: your new entry -> the new added entry?

packages/shopping/src/application.ts
bajtos: Well, what are the required fields? Isn't it just a `paths` object that can be empty?

```ts
this.api({
  components: {
    ...SECURITY_SCHEMA_SPEC,
  },
  security: SECURITY_SPEC,
  paths: {
    // will be filled by LoopBack from controller metadata
  }
});
```

jannyHou: @bajtos Not really, the problem is the `openapi` and `info` fields, they are required and especially the `info` object, you cannot provide an empty value like `''` or `{}`.

See my last commit, to call `this.api()`, I have to provide valid value for `openapi` and `info`, but the rest server doesn't overwrite them later.

The good spec filled by rest server:
Screen Shot 2019-09-16 at 11 04 53 AM

The one generated by calling `this.api()` with "empty" `openapi` and `info` fields:

Screen Shot 2019-09-16 at 11 08 58 AM

We can improve this by defining and checking the empty value of `openapi` and `info`, like
```ts
// in rest server
if (spec.openapi == '') {// generate openapi field }
if (spec.info.title == '' && spec.info.version == '') {// generate info field}
```

Thought? If this is the right direction for improvement I can create an issue (or just submit a PR)

bajtos: I see, thank you for explaining the details!

I think that ideally, I'd like:

- `app.api()` require fields like `info` to force the user to provide these required fields. The current implementation is good for that :+1:
- Have a way for extensions to contribute additional OpenAPI metadata. This could be a new method in `Application` or `RestServer`, but we can also use binding-based approach where the app collects OpenAPI metadata from bindings with a given tag (or other metadata-based specification). I think this is out of scope of your current work (both on the spike and on enabling auth in swagger-ui), so let's create a spike issue to research different options.

As for this spike PR, I think the current proposal (as shown in the committed code) is good enough and no further changes are necessary :+1: 

/cc @raymondfeng

packages/shopping/src/application.ts
```diff
@@ -47,6 +48,13 @@ export class ShoppingApplication extends BootMixin(
 ) {
   constructor(options?: ApplicationConfig) {
     super(options);
+    this.api({
+      openapi: '3.0.0',
+      info: {title: '', version: ''},
```
bajtos: Can you fill `title` and `version` with something more meaningful please? Personally, I'd use metadata from `package.json`.

```ts
import * as package from '../package.json';
// ...

info: {title: package.name, version: package.version}
```

packages/shopping/src/application.ts
raymondfeng: We're already exploring the idea to create an extension point to enhance the generated OpenAPI spec so that the application or other extensions can add code to transform the spec.

raymondfeng:
@jannyHou Please fix the [code-lint issue](https://travis-ci.com/strongloop/loopback4-example-shopping/jobs/235953482).

packages/shopping/src/application.ts
jannyHou: updated 👌

jannyHou:
Follow up stories see https://github.com/strongloop/loopback-next/issues/2027#issuecomment-531395515

packages/shopping/src/application.ts
```diff
@@ -47,6 +48,13 @@ export class ShoppingApplication extends BootMixin(
 ) {
   constructor(options?: ApplicationConfig) {
     super(options);
+    this.api({
+      openapi: '3.0.0',
+      info: {title: pkg.name, version: pkg.version},
+      paths: {},
```
jannyHou: servers: [{url: '/'}]

jannyHou: or config it when create the rest app:

https://loopback.io/doc/en/lb4/Server.html#customize-how-openapi-spec-is-served

jannyHou:
Closed as the tutorial is added :)

PR #301, opened 2019-10-03
feat: enable auth token dialog in api explorer

emonddr:
Make the authentication token dialog available in API Explorer

Connected to https://github.com/strongloop/loopback-next/issues/3740

### DIFF
--- a/packages/shopping/src/application.ts
+++ b/packages/shopping/src/application.ts
@@ -29,6 +29,7 @@ import {
 import {PasswordHasherBindings} from './keys';
 import {BcryptHasher} from './services/hash.password.bcryptjs';
 import {JWTAuthenticationStrategy} from './authentication-strategies/jwt-strategy';
+import {SECURITY_SCHEME_SPEC} from './utils/security-spec';
 
 /**
  * Information from package.json
@@ -47,6 +48,19 @@ export class ShoppingApplication extends BootMixin(
 ) {
   constructor(options?: ApplicationConfig) {
     super(options);
+
+    /*
+       This is a workaround until an extension point is introduced
+       allowing extensions to contribute to the OpenAPI specification
+       dynamically.
+    */
+    this.api({
+      openapi: '3.0.0',
+      info: {title: pkg.name, version: pkg.version},
+      paths: {},
+      components: {securitySchemes: SECURITY_SCHEME_SPEC},
+      servers: [{url: '/'}],
+    });
 
     this.setUpBindings();
 
--- a/packages/shopping/src/controllers/user.controller.ts
+++ b/packages/shopping/src/controllers/user.controller.ts
@@ -29,6 +29,7 @@ import {
   UserServiceBindings,
 } from '../keys';
 import * as _ from 'lodash';
+import {OPERATION_SECURITY_SPEC} from '../utils/security-spec';
 
 export class UserController {
   constructor(
@@ -102,6 +103,7 @@ export class UserController {
   }
 
   @get('/users/me', {
+    security: OPERATION_SECURITY_SPEC,
     responses: {
       '200': {
         description: 'The current user profile',
--- a/packages/shopping/src/utils/security-spec.ts
+++ b/packages/shopping/src/utils/security-spec.ts
@@ -1,0 +1,18 @@
+// Copyright IBM Corp. 2018,2019. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {SecuritySchemeObject, ReferenceObject} from '@loopback/openapi-v3';
+
+export const OPERATION_SECURITY_SPEC = [{bearerAuth: []}];
+export type SecuritySchemeObjects = {
+  [securityScheme: string]: SecuritySchemeObject | ReferenceObject;
+};
+export const SECURITY_SCHEME_SPEC: SecuritySchemeObjects = {
+  bearerAuth: {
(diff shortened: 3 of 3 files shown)


emonddr:
Hmm, I no longer see 

image

in the top left of API Explorer.

Comparing generated OpenAPI spec from original code, and OpenAPI spec generated by the code in my first commit, I see a missing section:

image

I will try the original approach @jannyHou  tried in the 1st commit of her PoC.

https://github.com/strongloop/loopback4-example-shopping/pull/267/commits/5f9b027c6b744825c9a9f6e5eb5986979fad7d7f

emonddr:
If we add securitySchemes to the open api spec post app.boot(), then the
`servers` information is preserved (not wiped out), and I see the 
`Servers` combo box AND the `Authorize` button at the top of the 
API Explorer

image

agnes512:
One of the criteria:
> Decide the scope of the endpoints that you want to secure with the bearer security schema. (The spike apply it globally, but the formal PR should decide the scope first)

Do we include the endpoint scope in this PR?

packages/shopping/src/application.ts
```diff
@@ -99,4 +101,15 @@ export class ShoppingApplication extends BootMixin(
 
     this.bind(UserServiceBindings.USER_SERVICE).toClass(MyUserService);
   }
+
+  modifyApiSpecAndRebind() {
+    let openAPISchema = this.getSync(RestBindings.API_SPEC);
+    addSecuritySchemesToOpenApiSpec(openAPISchema);
+    this.bind(RestBindings.API_SPEC).to(openAPISchema);
```
raymondfeng: No need to rebind if the update to open api spec is in place.

packages/shopping/src/application.ts
```diff
@@ -99,4 +101,15 @@ export class ShoppingApplication extends BootMixin(
...
+  }
+
+  async start(): Promise<void> {
+    this.modifyApiSpecAndRebind();
```
raymondfeng: We should add a comment claiming this is a workaround before an extension point for open api schema generation is introduced.

packages/shopping/src/application.ts
```diff
@@ -99,4 +101,15 @@ export class ShoppingApplication extends BootMixin(
 
     this.bind(UserServiceBindings.USER_SERVICE).toClass(MyUserService);
   }
+
+  modifyApiSpecAndRebind() {
```
raymondfeng: Let's rename it as `updateOpenApiSpecWithSecurity`.

packages/shopping/src/utils/security-spec.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright IBM Corp. 2018,2019. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {SecuritySchemeObject, ReferenceObject} from '@loopback/openapi-v3';
+import {OpenApiSpec} from '@loopback/rest';
+
+export const SECURITY_SPEC_OPERATION = [{bearerAuth: []}];
```
raymondfeng: `OPERATION_SECURITY_SPEC`?

emonddr:
The default OpenAPI schema created by [rest.component.ts](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.component.ts#L91
)


is:

{"openapi":"3.0.0","info":{"title":"LoopBack Application","version":"1.0.0"},"paths":{},"servers":[{"url":"/"}]}

and then [rest.server.ts](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts) adds to the "paths" portion and the "component.schemas" portion of the OpenAPI spec.

emonddr:
> One of the criteria:
> 
> > Decide the scope of the endpoints that you want to secure with the bearer security schema. (The spike apply it globally, but the formal PR should decide the scope first)
> 
> Do we include the endpoint scope in this PR?

The original shopping cart example only secured 1 endpoint : `/users/me` .
And to send the JWT token to that endpoint, the user had to use: a) curl command, or b) another REST Client like Postman. Now that the API Explorer has the capability to set and use a JWT token, the user can perform this in the API Explorer completely.
To avoid changing the sample too much, we will only specify the security requirement object setting on the `/users/me` endpoint.

There are other changes coming to the sample... adding authorization on top of the authentication. Any major changes should happen in that work. :)

packages/shopping/src/application.ts
emonddr: This code will be obsolete in my next commit; based on our conversation yesterday.

packages/shopping/src/application.ts
emonddr: This code will be obsolete in my next commit; based on our conversation yesterday.

packages/shopping/src/application.ts
emonddr: This code will be obsolete in my next commit; based on our conversation yesterday. 
But I definitely will provide a comment above the `this.api()` call. :)

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -41,6 +50,7 @@ export class UserController {
     public jwtService: TokenService,
     @inject(UserServiceBindings.USER_SERVICE)
     public userService: UserService<User, Credentials>,
+    @inject(RestBindings.Http.REQUEST) private request: Request,
```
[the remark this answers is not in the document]
emonddr: Thanks, missed that one. I would've thought our build process would catch unused variables.

PR #311, opened 2019-10-09
chore: relax tests to avoid CI failures

raymondfeng:
See https://travis-ci.com/strongloop/loopback4-example-shopping/jobs/243828270


emonddr:
> See https://travis-ci.com/strongloop/loopback4-example-shopping/jobs/243828270 

I was wondering why that test was failing.

Thanks for the fix.

PR #268, opened 2019-09-09
feat: deploy shopping example as a Kubernetes cluster

raymondfeng:
This PR represents an experiment to decompose the shopping app into multiple microservices, each of which is packaged as a docker image. It's possible to communicate between microservices usign REST and/or gRPC.

A `helm` chart is added as `kubernetes/shopping-app` so that we can use `helm` to install the application. Please note it depends on `mongodb` and `redis` charts.

See https://github.com/strongloop/loopback4-example-shopping/blob/microservices/kubernetes/create-k8s.sh for instructions to get the application up and running with minikube.

diagram

PR #333, opened 2019-10-23
feat: add grpc support for recommender service

raymondfeng:
This is a spin-off from https://github.com/strongloop/loopback4-example-shopping/pull/268.

1. Allow recommender service to expose gRPC endpoints
2. Allow shopping to connect to recommender over gRPC and REST

### DIFF
--- a/packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
+++ b/packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
@@ -11,11 +11,14 @@ import {setupApplication} from './helper';
 import {
   createRecommendationServer,
   HttpServer,
+  createGRPCRecommendationServer,
 } from 'loopback4-example-recommender';
 import {PasswordHasher} from '../../services/hash.password.bcryptjs';
 import {PasswordHasherBindings, TokenServiceConstants} from '../../keys';
 import {JWTService} from '../../services/jwt-service';
 import {securityId} from '@loopback/security';
+import {RecommenderGrpc} from '../../services';
+import {Server} from 'grpc';
 
 const recommendations = require('loopback4-example-recommender/data/recommendations.json');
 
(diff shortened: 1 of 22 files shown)


packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -238,21 +241,33 @@ describe('UserController', () => {
...
         .get(`/users/${newUser.id}/recommend`)
-        .expect(200, recommendations);
+        .expect(200, recommendations['user001']);
+    });
+
+    it('returns product recommendations for a user using gRPC', async () => {
+      const recommender = await app.get<RecommenderService>(
+        'services.RecommenderGrpcService',
+      );
+      const products = await recommender.getProductRecommendations('user001');
```
agnes512: Can we have a test case for the user not in recommendations such as `user003`?

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -238,21 +241,33 @@ describe('UserController', () => {
 
   describe('user product recommendation (service) api', () => {
     let recommendationService: HttpServer;
+    let recommendationGRPCService: Server;
 
     before(async () => {
       recommendationService = createRecommendationServer();
       await recommendationService.start();
+      recommendationGRPCService = createGRPCRecommendationServer();
+      recommendationGRPCService.start();
```
agnes512: Do we need `await` for this line as well?

raymondfeng: No. This start does not return a promise.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
raymondfeng: The recommender is a mock service that returns same products for users other than `user002`

raymondfeng: I added one for user004.

PR #338, opened 2019-10-24
feat: tidy up tests and make datasources compatible with kubernetes deployment

raymondfeng:
Spin-off from https://github.com/strongloop/loopback4-example-shopping/pull/268

1. Improve tests to avoid side-efforts (make sure setup code to execute in mocha hooks)
2. Make sure datasources are disconnected upon `app.stop` 
3. Add logic to `*.datasources.ts` to accept configuration from env vars by kubernetes


packages/shopping/src/__tests__/acceptance/authentication.acceptance.ts
```diff
@@ -196,7 +200,8 @@ describe('authentication services', () => {
   });
 
   async function setupApp() {
-    app = await setupApplication();
+    const appWithClient = await setupApplication();
```
emonddr: @raymondfeng , how were these tests working before if the wrong object was being assigned to `app`?

bin/rebuild-package-locks.js
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env node
```
emonddr: @raymondfeng , why do we need to rebuild package locks?
Is this a script we should make available to anyone building an LB4 app?

packages/shopping/src/datasources/mongo.datasource.ts
```diff
@@ -3,17 +3,34 @@
...
 
+function updateConfig(dsConfig: AnyObject) {
+  if (process.env.KUBERNETES_SERVICE_HOST) {
+    dsConfig.host = process.env.SHOPPING_APP_MONGODB_SERVICE_HOST;
+    dsConfig.port = +process.env.SHOPPING_APP_MONGODB_SERVICE_PORT!;
+  }
+  return dsConfig;
+}
+
+@lifeCycleObserver('datasource')
```
emonddr: Very cool use of life cycle observers.

packages/shopping/src/__tests__/acceptance/authentication.acceptance.ts
raymondfeng: Haha, the test was moved from `unit` to `acceptance` and it used to have a different `helper`.

packages/shopping/src/datasources/mongo.datasource.ts
raymondfeng: We now generate datasources in this pattern with `lb4 datasource`.

bin/rebuild-package-locks.js
raymondfeng: The shopping project is set up as a lerna monorepo and it requires a few steps to rebuild package-lock files.

PR #356, opened 2019-10-31
chore: remove docker:start and docker:stop as they are now part of mocha hooks

raymondfeng:
The start/stop of docker containers are now handled in Mocha before/after hooks with testcontainers. See https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/__tests__/fixtures/redis-mongo-docker.ts.


emonddr:
@raymondfeng , I don't think we should remove these. What if I am not running `npm test`, and I want to `try out` the application? e.g. https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#try-it-out   , uses `docker:start`

raymondfeng:
@emonddr Good catch. I reverted the change to root package.json. PTAL.

PR #385, opened 2019-11-14
feat: extract local credentials into a new model

bajtos:
Introduce `UserCredentials` models to hold hashed passwords, add has-one relation from `User` to `UserCredentials`.

Rework authentication-related code to work with the new domain model.

See https://github.com/strongloop/loopback-next/issues/1996 for additional information.

_I have to say the code in the example app is a bit messy. Some things are (unnecessarily) duplicated in multiple places, therefore my pull request had to touch quite few files to update all duplications. Considering our current priorities, I am leaving refactoring & cleanup out of scope of my PR._

### DIFF
--- a/packages/shopping/src/controllers/user.controller.ts
+++ b/packages/shopping/src/controllers/user.controller.ts
@@ -3,9 +3,16 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {repository} from '@loopback/repository';
+import {repository, model, property} from '@loopback/repository';
 import {validateCredentials} from '../services/validator';
-import {post, param, get, requestBody, HttpErrors} from '@loopback/rest';
+import {
+  post,
+  param,
+  get,
+  requestBody,
+  HttpErrors,
+  getModelSchemaRef,
+} from '@loopback/rest';
 import {User, Product} from '../models';
 import {UserRepository} from '../repositories';
 import {RecommenderService} from '../services/recommender.service';
@@ -30,6 +37,15 @@ import {
 } from '../keys';
 import * as _ from 'lodash';
 import {OPERATION_SECURITY_SPEC} from '../utils/security-spec';
+
+@model()
+export class NewUserRequest extends User {
+  @property({
+    type: 'string',
+    required: true,
+  })
+  password: string;
+}
 
 export class UserController {
   constructor(
@@ -58,18 +74,36 @@ export class UserController {
       },
     },
   })
-  async create(@requestBody() user: User): Promise<User> {
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(NewUserRequest, {
+            title: 'NewUser',
+          }),
+        },
+      },
+    })
+    newUserRequest: NewUserRequest,
+  ): Promise<User> {
     // ensure a valid email value and password value
-    validateCredentials(_.pick(user, ['email', 'password']));
+    validateCredentials(_.pick(newUserRequest, ['email', 'password']));
 
     // encrypt the password
-    // eslint-disable-next-line require-atomic-updates
-    user.password = await this.passwordHasher.hashPassword(user.password);
+    const password = await this.passwordHasher.hashPassword(
+      newUserRequest.password,
+    );
 
     try {
       // create the new user
-      const savedUser = await this.userRepository.create(user);
-      delete savedUser.password;
+      const savedUser = await this.userRepository.create(
+        _.omit(newUserRequest, 'password'),
+      );
+
+      // set the password
+      await this.userRepository
+        .userCredentials(savedUser.id)
+        .create({password});
 
       return savedUser;
     } catch (error) {
@@ -97,9 +131,7 @@ export class UserController {
     },
   })
   async findById(@param.path.string('userId') userId: string): Promise<User> {
-    return this.userRepository.findById(userId, {
-      fields: {password: false},
-    });
+    return this.userRepository.findById(userId);
   }
 
   @get('/users/me', {
--- a/packages/shopping/src/models/user-credentials.model.ts
+++ b/packages/shopping/src/models/user-credentials.model.ts
@@ -1,0 +1,40 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class UserCredentials extends Entity {
+  @property({
+    type: 'string',
+    id: true,
+    mongodb: {dataType: 'ObjectID'},
+  })
+  id: string;
+
+  @property({
+    type: 'string',
+    required: true,
+  })
+  password: string;
+
+  @property({
+    type: 'string',
+    required: true,
+    mongodb: {dataType: 'ObjectID'},
+  })
+  userId: string;
+
+  constructor(data?: Partial<UserCredentials>) {
+    super(data);
+  }
+}
+
+export interface UserCredentialsRelations {
+  // describe navigational properties here
+}
+
+export type UserCredentialsWithRelations = UserCredentials &
+  UserCredentialsRelations;
--- a/packages/shopping/src/repositories/user-credentials.repository.ts
+++ b/packages/shopping/src/repositories/user-credentials.repository.ts
@@ -1,0 +1,18 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {DefaultCrudRepository, juggler} from '@loopback/repository';
+import {UserCredentials, UserCredentialsRelations} from '../models';
+import {inject} from '@loopback/core';
+
+export class UserCredentialsRepository extends DefaultCrudRepository<
+  UserCredentials,
+  typeof UserCredentials.prototype.id,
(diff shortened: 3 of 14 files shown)


packages/shopping/src/controllers/user.controller.ts
```diff
@@ -58,18 +74,37 @@ export class UserController {
...
     },
   })
-  async create(@requestBody() user: User): Promise<User> {
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(NewUserRequest, {
+            title: 'NewUser',
+            exclude: ['id'],
```
emonddr: Why are we excluding id?

emonddr: In the API explorer, we had to choice of passing or not passing an `id` value when creating a user

packages/shopping/src/repositories/user-credentials.repository.ts
```diff
@@ -0,0 +1,14 @@
...
+import {UserCredentials, UserCredentialsRelations} from '../models';
+import {MongoDataSource} from '../datasources';
+import {inject} from '@loopback/core';
+
+export class UserCredentialsRepository extends DefaultCrudRepository<
+  UserCredentials,
+  typeof UserCredentials.prototype.id,
+  UserCredentialsRelations
+> {
+  constructor(@inject('datasources.mongo') dataSource: MongoDataSource) {
```
emonddr: Why MongoDataSource? UserRepository uses juggler.DataSource 

https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/repositories/user.repository.ts#L28

packages/shopping/src/controllers/user.controller.ts
bajtos: Good point.

In a typical application, the id is generated by the database. That's how we are scaffolding new models via `lb4 model` and `lb4 controller` too. 

I'll double check what's the actual setup in this example app and ensure the controller is consistent with model.

packages/shopping/src/repositories/user-credentials.repository.ts
bajtos: I used `lb4 repository` to create this file 🤷‍♂ 

I am fine to use `juggler.DataSource` for consistency 👍 

Later, we can rework all repositories to use the style scaffolded by `lb4 repository`, but I feel that's out of scope of this pull request.

packages/shopping/src/models/user-credentials.model.ts
```diff
@@ -0,0 +1,40 @@
...
+    required: true,
+  })
+  password: string;
+
+  @property({
+    type: 'string',
+    required: true,
+    mongodb: {dataType: 'ObjectID'},
+  })
+  userId: string;
```
jannyHou: Hmm, why do we need an extra userId when `id` is already defined?

packages/shopping/src/controllers/user.controller.ts
jannyHou: redis keeps failing on my local I couldn't try creating a user with explorer, but `forceId` should be default to `true`, so I think excluding the id field is correct here.

bajtos: I believe that `forceId` is enabled only when the `id` property is defined with `generated: true`, which is not the case in our Shopping service - see here:

https://github.com/strongloop/loopback4-example-shopping/blob/19d6c9265e46eb185c8600617134835985797082/packages/shopping/src/models/user.model.ts#L25-L29

I am going to remove `exclude: ['id']` for consistency with other parts of the codebase. We can revisit this aspect later, it's out of scope of my pull request.

packages/shopping/src/models/user-credentials.model.ts
bajtos: - `id` is the primary key of `UserCredentials` model
- `userId` is a foreign key referencing id of the user these credentials belong to

This is the classic HasOne/BelongsTo setup we use everywhere else in LB4.

It is possible to use User's id as the primary key of UserCredentials too, but we don't have first-class support for that in LoopBack yet.

Also note that depending on the password policy, we may want to store historical credentials and thus have more than on UserCredentials instance for a single user. Think of a password policy "the new value must be different than the last 3 values used".

bajtos:
@emonddr @jannyHou thank you for thoughtful comments. I have addressed some of them in 24ff52c and posted comments where I think no action is needed.

PTAL, LGTY now?

It makes me wonder, is there anything else we need to update as part of this change? E.g. documentation?

packages/shopping/src/controllers/user.controller.ts
emonddr: In the shopping app, is has always been historically possible to enter an id for a new User when posting to /users. And it has historically been possible to avoid specifying an id (so that it is automatically generated) when creating a user. Let's make sure this still works. So let's be sure not to modify any properties if the id field. thx. Please confirm this still works.

emonddr: @bajtos ^

emonddr:
> @emonddr @jannyHou thank you for thoughtful comments. I have addressed some of them in [24ff52c](https://github.com/strongloop/loopback4-example-shopping/commit/24ff52c09602880d2ccfa4334ca55be6c1b138ed) and posted comments where I think no action is needed.
> 
> PTAL, LGTY now?
> 
> It makes me wonder, is there anything else we need to update as part of this change? E.g. documentation?

Please look at https://loopback.io/doc/en/lb4/Authentication-Tutorial.html , and double-check that any code examples shown or instructions to post or get a user match what you now have. 

The `Try it out` section , https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#try-it-out  passes an id of "1" in the POST /users .   This doesn't work in your PR now, please make sure a user is able to POST /users  where  id value is specified, and where id is not specified but autogenerated by mongo db.

packages/shopping/src/controllers/user.controller.ts
emonddr: > redis keeps failing on my local I couldn't try creating a user with explorer, but `forceId` should be default to `true`, so I think

that is weird.

You may need to run:

docker volume prune
docker images prune
docker prune

to clean up docker.

Usually if a container isn't running when it should.... a clean up like this fixes it.

bajtos: > In the shopping app, is has always been historically possible to enter an id for a new User when posting to /users. And it has historically been possible to avoid specifying an id (so that it is automatically generated) when creating a user. Let's make sure this still works. So let's be sure not to modify any properties of the id field. thx. Please confirm this still works.

All existing tests are passing. AFAICT, they are usually not providing any `id` value and relying on the database to generate one.

As for providing a custom id value, either it's covered by tests and then it is still working (because all tests are passing), or else the behavior is undefined as far as I am concerned in this pull request 🤷‍♂

bajtos:
> The Try it out section , https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#try-it-out passes an id of "1" in the POST /users . This doesn't work in your PR now, please make sure a user is able to POST /users where id value is specified, and where id is not specified but autogenerated by mongo db.

I find it very annoying that the documentation is depending on behavior that's not covered by automated tests :-/

I'll take a look and add the missing test.

bajtos:
Thank you @emonddr for the pointers for documentation which I should check, this was very helpful 🙇

jannyHou:
@bajtos Thank you, 
> It makes me wonder, is there anything else we need to update as part of this change? E.g. documentation?


You probably want to explain the `UserCredential` is separated from `User` as a single model in https://github.com/strongloop/loopback4-example-shopping#models

packages/shopping/src/controllers/user.controller.ts
bajtos: Please ignore my comment above, I'll investigate this use case based on what we have in our documentation, as discussed elsewhere in this PR.

bajtos:
Thank you all again for your feedback. I see the following major areas to improve:

- Ensure we can provide a custom id when creating a new user, update docs and tests accordingly
- Update "Authentication tutorial" to match the new example app


Here is what I did:

- In https://github.com/strongloop/loopback-next/pull/4201, I changed the example data in "Try it out" section to use a valid ObjectID value instead of `1`
- In this pull request (e27afa7), I added a new test to verify that it's possible to create a new user using the payload shown in our docs
- In https://github.com/strongloop/loopback-next/pull/4201, I updated the relevant parts of "Authentication Tutorial" to match the updates made by this pull request
- Finally, in this pull request (e27afa7), I added `UserCredentials` model to README.

@emonddr @jannyHou LGTY now?

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -58,18 +74,36 @@ export class UserController {
       },
     },
   })
-  async create(@requestBody() user: User): Promise<User> {
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(NewUserRequest, {
+            title: 'NewUser',
```
jannyHou: Should we exclude 'id' in the schema to be consistent with the type? The type for data is `newUserRequest: Omit<NewUserRequest, 'id'>,`

bajtos: Good catch.

We must allow `id` in the payload, see the test below and also the docs. I'll update the type to `newUserRequest: NewUserRequest`.

emonddr:
> In strongloop/loopback-next#4201, I changed the example data in "Try it out" section to use a valid ObjectID value instead of 1 

As I mentioned in the PR referenced above, if we are not going to pass in an `easy` id like  1, 2, or 3, etc but an Object ID (which is harder for a person to think of), let's just avoid passing an `id` value when POST-ing a new user in the `Try it Out` section of the document.

Issue #425, opened 2019-12-09
Website for Shopping App


hacksparrow:
In the Shopping App Brainstorming session last week, it was decided that we will upgrade [loopback4-example-shopping ](https://github.com/strongloop/loopback4-example-shopping) to have a corresponding website.

**Acceptance Criteria**

- [x] [Authentication and authorization clean up](https://github.com/strongloop/loopback4-example-shopping/issues/432)
- [x] [CRUD operations on shopping carts](https://github.com/strongloop/loopback4-example-shopping/issues/475)
- [ ] [CRUD operations on orders](https://github.com/strongloop/loopback4-example-shopping/issues/474)
- [x] [View products](https://github.com/strongloop/loopback4-example-shopping/issues/439)
- [ ] [CRUD operation on products](https://github.com/strongloop/loopback4-example-shopping/issues/440)
- [x] [Order products](https://github.com/strongloop/loopback4-example-shopping/issues/441)
- [ ] [Product recommendation](https://github.com/strongloop/loopback4-example-shopping/issues/442)
- [ ] [Persona](https://github.com/strongloop/loopback4-example-shopping/issues/443)
- [ ] [UI testing](https://github.com/strongloop/loopback4-example-shopping/issues/476)
- [ ] [Documentation](https://github.com/strongloop/loopback4-example-shopping/issues/477)
- [ ] [Misc.](https://github.com/strongloop/loopback4-example-shopping/issues/478)

dhmlau:
Let's move it to the `loopback4-example-shopping` because we'll be implementing the frontend in here.

Issue #439, opened 2019-12-18
Shopping App Website: View products


hacksparrow:
Part of https://github.com/strongloop/loopback4-example-shopping/issues/425.

**Acceptance Criteria**

- [x] Basic infrastructure of the web site
- [x] 15 pre-populated products
- [x] Show products on landing page
- [x] Click on a product to see the details

dhmlau:
Since this is already part of PR https://github.com/strongloop/loopback4-example-shopping/pull/468, I'm including this in `2020Q1`.

dhmlau:
@hacksparrow, since PR https://github.com/strongloop/loopback4-example-shopping/pull/468 has landed, I believe this issue is done? If yes, could you please close it. Thanks.

dhmlau:
Closing as done.

Issue #432, opened 2019-12-15
Authentication and Authorization clean up


dhmlau:
The authentication and authorization example in the code is confusing and leads to many questions. Update the example to:

- [x] Get rid of CASBIN and show a custom implementation of authorization rules
- [x] Use the user id as the unique identifier while enforcing authorization 
- [x] Update the corresponding tests

---
ORIGINAL DESCRIPTION

In the discussion with @emonddr, it seems like we don't have a seamless flow for getting users authenticated and such users (with the same userid) are being authorized for certain actions. 

## Proposal 
At the high level, we should support the following scenarios, reusing our fictitious users `Alice` and `Bob` as listed in https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/fixtures/casbin/rbac_policy.csv:
- [x] Login as a regular user `Alice`, she can find and create orders. 
- [x] Login as a customer service rep `Bob`, he can find, create and delete orders
- [x] Login as a customer service supervisor `Clara`, she can find, create, update, patch and delete orders

Besides the pre-existing users, the shopping app allows to sign up as a customer.  The person then will have the same access as `Alice` (the regular user).

@emonddr, please chime in if I miss anything. Thanks. 

## Acceptance Criteria
- [ ] The CASBIN CSV file should contain a better  `user identifier` than : bob, alice . Perhaps the email?
- [ ] All authorizers use a consistent way of figuring out what user we are dealing with. Is it 'id` or `name` or `email` field. Does it make sense they don't use the same field to reject or allow access? If so, document clearly why this is so.
- [ ] Each authenticated user can only perform the actions on the resources for which they are authorized. This needs to be tested manually in API Explorer, and via mocha tests.

emonddr:
The current obstacle is that the current shopping example app uses a specific user model
https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/models/user.model.ts#L24 
which has id, firstName,lastName, email, password etc.

The user logs in with email id and password.  So the uniqueness of user is determined by email.


Before a JWT token is created, we convert a User to UserProfile

https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/services/user-service.ts#L51 


```
convertToUserProfile(user: User): UserProfile {
    // since first name and lastName are optional, no error is thrown if not provided
    let userName = '';
    if (user.firstName) userName = `${user.firstName}`;
    if (user.lastName)
      userName = user.firstName
[...]

PR #468, opened 2020-01-08
Website for Shopping App: UI and basic functionality

hacksparrow:
Part of https://github.com/strongloop/loopback4-example-shopping/issues/425.

I have been working on a website interface for the API in the example shopping app. To see the website, clone https://github.com/strongloop/loopback4-example-shopping and checkout out the website branch. Start the app, run `npm run migrate`  and then load http://localhost:3000/shoppy.html.

Apart from creating the website, this is also an exercise in cleaning up how authorization is handled in the app and applying the necessary protections to the API endpoints.

There's some more work to be done, but would like to hear your feedback on the progress so far.


emonddr:
Great start, Yaapa!

emonddr:
After removing item from cart, the menu item for the cart still shows a number. I click it again, and the
item is still there. A refresh is needed on the page (after removal from cart) to show a proper removal.

emonddr:
As a regular user, it is not clear what the `Search` area is for. Perhaps a tooltip?

As an Admin, I thought I would be able to accomplish more tasks on the main
screen. Currently it is the same as other users. I realize this would change in time.
It is also not clear what the `Search` area is for an admin. Also, should an Admin be allowed to shop?

Just some initial thoughts.

emonddr:
Since our lb4 server uses Typescript, shouldn't our front-end use Typescript as well?

emonddr:
I find it confusing that I am presented `twice` with an `Add to Cart` button.
image

Initially I was looking at the code in ui.js `addToCart` and found the logic for $('#addToCart').show(); and $('#addToCart').hide(); confusing which led me to take a look at the UI behaviour again.

emonddr:
Perhaps have an add to cart button which `silently` updates the shopping cart. If user wants to see the contents of the shopping cart, they go to the shopping cart to view the contents?

emonddr:
`Qauntity` => Quantity

emonddr:
In `api.js` for this method:
`getProducts(cb, skip = 0, limit = 4)`
please put callback `cb` at the end to conform to other functions in same file. thx.

emonddr:
In `product.js` why are product.details obtained from the YAML instead of the product object
returned from the database call?

```js
api.getProduct(id, function (product) {
      if (product) {
        const productHtml = productTemplate.replace(/#NAME#/ig, product.name)
                            .replace(/#PRICE#/ig, new Intl.NumberFormat('en-US', { maximumSignificantDigits: 3 }).format(product.price))
                            .replace(/#UNFORMATTED-PRICE#/g, product.price)
                            .replace(/#IMAGE#/ig, product.image)
                            .replace(/#DESCRIPTION#/ig, product.description)
                            .replace(/#DETAILS#/ig, parseYml(product.details))
                            .replace(/#ID#/ig, product.productId);
        $('#product').append(productHtml);
        $('#product').append(addToCartTemplate);
      }
    });
```

packages/shopping/public/api.js
```diff
@@ -0,0 +1,102 @@
...
+      error: errCb
+    });
+  },
+
+  getShoppingCartItems(successCb, errCb) {
+    const userId = localStorage.getItem('shoppyUserId');
+    const token = localStorage.getItem('shoppyToken');
+    if (userId) {
+      const url = apiUrl + '/shoppingCarts/' + userId;
+      $.ajax({
```
bajtos: I believe all modern browsers support [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) by now, see [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility). Can you use that API instead of jQuery AJAX?

Similarly, we can use [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) instead of callbacks, see [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility).

---

For longer term, it would be great to leverage OpenAPI spec provided by the server and use either a code-generated client (see https://github.com/OpenAPITools/openapi-generator) or at least a dynamic client https://github.com/swagger-api/swagger-js. This part is of scope of this pull request though.

packages/shopping/public/shoppy.js
```diff
@@ -0,0 +1,35 @@
...
+  api.getProducts(function (products) {
+    if (products) {
+      products.forEach(product => {
+        const productHtml = productTemplate.replace(/#NAME#/ig, product.name)
+                            .replace(/#PRICE#/ig, new Intl.NumberFormat('en-US', { maximumSignificantDigits: 3 }).format(product.price))
+                            .replace(/#UNFORMATTED-PRICE#/g, product.price)
+                            .replace(/#IMAGE#/ig, product.image)
+                            .replace(/#DESCRIPTION#/ig, product.description)
+                            .replace(/#DETAILS#/ig, product.details)
+                            .replace(/#ID#/ig, product.productId);
```
bajtos: Have you considered using ES6 template literals? (See [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Browser_compatibility).)

```ts
const productHtml = `<div id="card-${product.productId}" class="card" style="width: 18rem;">
   <img src="${product.image}" class="card-img-top product-img" alt="${product.name}">
   (etc.)
 </div>
`;
```

More importantly, please remember to HTML-escape values when building HTML strings manually, we don't want our example application to be vulnerable to [cross-site scripting](https://www.owasp.org/index.php/Testing_for_Cross_site_scripting). Consider what happens when e.g. `product.description` contains the string `<script>window.alert("you have been hacked!");</script>`.

I think the easiest option is to use Handlebars for templating, see https://handlebarsjs.com. You can add the library to your front-page via a public CDN and it will handle tricky parts of HTML escaping for you.

If we had a webpack/parcel/browserify build pipeline for front-end assets in place, then I would recommend to get rid of jQuery and use something like https://www.npmjs.com/package/nanohtml

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -1,331 +1,331 @@
...
-      '-1',
-    );
-    const userProfile = {
-      [securityId]: newUser.id,
-      name: `${newUser.firstName} ${newUser.lastName}`,
-    };
-    expiredToken = await tokenService.generateToken(userProfile);
-  }
-});
+// // Copyright IBM Corp. 2018,2019. All Rights Reserved.
```
bajtos: Why is this acceptance test commented out?

packages/shopping/src/models/product.model.ts
```diff
@@ -1,22 +1,17 @@
-// Copyright IBM Corp. 2018. All Rights Reserved.
-// Node module: loopback4-example-shopping
-// This file is licensed under the MIT License.
-// License text available at https://opensource.org/licenses/MIT
```
bajtos: Please preserve the license boilerplate header.

packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
hacksparrow: There was a build issue. Uncommented.

packages/shopping/src/models/product.model.ts
hacksparrow: 👍

packages/shopping/public/shoppy.js
hacksparrow: All great feedback!

For this PR I want to keep things simple and focus on making things works. There will be another PR after this which will handle authorization. In the third PR, let's optimize UI implementation details, tooling, and testing (https://webdriver.io/).

hacksparrow:
@emonddr 
> In `product.js` why are product.details obtained from the YAML
There are not fetched from the YAML. `api.getProduct()` makes a request to the LB server.

> please put callback cb at the end to conform to other functions in same file. thx.
Made it consistent with others.

> Perhaps have an add to cart button which silently updates the shopping cart.
Labelled the button "Confirm". 

> Since our lb4 server uses Typescript, shouldn't our front-end use Typescript as well?
No. Let's not force TS on the front-end.

> As a regular user, it is not clear what the Search area is for.
For searching products. Feature not implemented yet.

> As an Admin, I thought I would be able to accomplish more tasks on the main
Coming up next week.

> After removing item from cart, the menu item for the cart still shows a number. 
Not able to reproduce. A screen capture would greatly help.

emonddr:
> screen capture would help...

image

Here there is this item in the cart. It shows (1). When I remove it via this button, the Shopping Cart menu item in the navbar will still say (1) , even though it is empty.

dhmlau:
> As an Admin, I thought I would be able to accomplish more tasks on the main
> Coming up next week.

I'm thinking that it might be better to keep the PR smaller in scope.  It seems like we have authentication working with the frontend, how about we have a separate PR for authorization clean + corresponding UI changes?

hacksparrow:
> I'm thinking that it might be better to keep the PR smaller in scope. It seems like we have authentication working with the frontend, how about we have a separate PR for authorization clean + corresponding UI changes?

Yes, Diana. Will update this PR's details for UI and the basic functionality.

packages/shopping/public/shoppy.js
bajtos: Fair enough. Please create a list of follow-up tasks so that we don't accidentally forget to address parts of my feedback provided in this pull request.

To be clear, a list of follow-up tasks does not have to be a series of GitHub issues, a checklist in a single GH issue may be a better solution.

> https://webdriver.io/

I am not very familiar with front-end testing tools, but I heard a lot of good feedback on
https://www.cypress.io. You may want to give it a try.

.eslintrc.js
```diff
@@ -5,4 +5,7 @@
 
 module.exports = {
   extends: ['@loopback/eslint-config'],
+  rules: {
+    '@typescript-eslint/no-unused-vars': 'off',
```
bajtos: I am very reluctant to disable this rule for the entire example codebase. If it's too much work to fix your front-end code to not rely on unused variables, then please disable this rule selectively for the front-end files only. 

See how we are using `overrides` config section in our "master" eslint config here:

https://github.com/strongloop/loopback-next/blob/3e15bcada2aa4f4cf6c0440b54d0233880b32c03/packages/eslint-config/eslintrc.js#L158-L166
```
  overrides: [
    {
      files: ['**/*.js'],
      rules: {
        '@typescript-eslint/prefer-optional-chain': 'off',
        '@typescript-eslint/prefer-nullish-coalescing': 'off',
      },
    },
  ],
```

packages/shopping/src/application.ts
```diff
@@ -127,4 +145,41 @@ export class ShoppingApplication extends BootMixin(
...
+    const userRepo = await this.getRepository(UserRepository);
+    await userRepo.deleteAll();
+    const usersDir = path.join(__dirname, '../fixtures/users');
+    const userFiles = fs.readdirSync(usersDir);
+
+    for (const file of userFiles) {
+      if (file.endsWith('.yml')) {
+        const userFile = path.join(usersDir, file);
+        const input = new NewUser(YAML.load(userFile));
+        const password = await passwordHasher.hashPassword(input.password);
```
bajtos: I am seeing this code pattern duplicated in quite few places now - in our controllers, acceptance tests and now in the migration script too.

IMO, it's time to enhance our user service and add a function for adding new users, one that can be used from all places. It's out of scope of this pull request, but please consider adding this task on your checklist.

packages/shopping/public/shoppy.js
hacksparrow: Sure!

bajtos: >  In the third PR, let's optimize UI implementation details, tooling, and testing

Well, I don't consider security vulnerabilities as "UI implementation details", but I won't block landing of this pull request on that 🤷‍♂

packages/shopping/public/product.html
```diff
@@ -0,0 +1,28 @@
...
+  <head>
+    <title>
+    Shoppy: The #1 online shopping destination
+    </title>
+    <link rel="icon" type="image/png" href="/favicon.png" />
+    <link href="https://fonts.googleapis.com/css?family=Pacifico&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="/style.css">
+    <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css" integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js" integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
```
dhmlau: For performance reasons, I was thinking it's better to put the `<script>` tag at the bottom of the `<body>`.  However, I just looked up and there seems to be another way, i.e. to use `async` or `defer`, see https://stackoverflow.com/questions/436411/where-should-i-put-script-tags-in-html-markup.

dougal83:
> I share the same view. In the "more modern" UI frameworks, such as Angular and React, they provide an easy way to call the backend APIs, and also have the UI elements come with it (Angular and React have, I don't know much about Vue.js to comment on that). I'd refer to use one of those too. I'm ok to make this PR land in this form and work on the next iteration later.

If people have the time, it wouldn't hurt to have several flavours of front end examples with this repo being the basis.  Angular /w Ngrx would be spiffing.

hacksparrow:
>  In the "more modern" UI frameworks, such as Angular and React ...

I am actually not in favor of using any framework. Reasons:

1. People have strong preferences for frameworks. That would alienate users who don't use or like the framework we use in the example.
2. Frameworks don't provide any easier API than `fetch`. What they excel at is state management. Our example app is going to be pretty simple, we don't need a framework.
3. Updating the framework with new releases is extra work.
4. The current plain JS implementation and simple and straightforward enough for anyone to port the functionality to any framework of their choice.

Let's keep the example app framework-neutral.

packages/shopping/public/product.html
hacksparrow: I have put the scripts at the bottom of the body tags. Can't use `async` or `defer`, since the order of loaded files dictate dependency.

hacksparrow:
@bajtos I have update eslintconfig so that @typescript-eslint/no-unused-vars is disabled only for the new front-end scripts. Also, running new DOM text through DOMPurify to prevent XSS.

packages/shopping/public/shoppy.js
hacksparrow: @bajtos follow up tasks - https://github.com/strongloop/loopback4-example-shopping/issues/478.

Added DOMPurity for sanitizing input.

packages/shopping/public/api.js
hacksparrow: > For longer term, it would be great to leverage OpenAPI spec provided 

Opened this issues https://github.com/strongloop/loopback4-example-shopping/issues/479.

packages/shopping/src/application.ts
hacksparrow: Follow up task https://github.com/strongloop/loopback4-example-shopping/issues/480.

dhmlau:
Related to:
- #432: authentication portion so far
- #439: basic infrastructure + view products
- #475: shopping cart

packages/shopping/public/templates.js
```diff
@@ -0,0 +1,221 @@
...
+            <a class="nav-link" href="/shoppy.html">Home <span class="sr-only">(current)</span></a>
+          </li>
+          <li class="nav-item" id="shoppingCartLink" style="display:none">
+            <a class="nav-link" href="javascript:displayShoppingCart()">Shopping Cart<span style="display:none" id="itemsInCart" class="badge badge-notify">0</span></a>
+          </li>
+          <li id="signUp" class="nav-item">
+            <a class="nav-link" href="#" data-toggle="modal" data-target="#signUpModal" tabindex="-1">Sign Up</a>
+          </li>
+
+<div class="modal" id="signUpModal" tabindex="-1" role="dialog">
```
dhmlau: nitpick: should this part be indented (for readability)?

packages/shopping/package.json
```diff
@@ -62,7 +63,8 @@
     "loopback-connector-grpc": "1.3.0",
     "loopback-connector-kv-redis": "3.0.2",
     "loopback-connector-mongodb": "5.2.1",
-    "loopback-connector-rest": "3.6.0"
+    "loopback-connector-rest": "3.6.0",
+    "yaml": "^1.7.2"
```
raymondfeng: Does `@types/yamljs` match `yaml`?

packages/shopping/public/purify.js
```diff
@@ -0,0 +1,2 @@
...ply(t,n)});var v=Object.seal||function(e){return e},b=v(/\{\{[\s\S]*|[\s\S]*\}\}/gm),T=v(/<%[\s\S]*|[\s\S]*%>/gm),A=v(/^data-[\-\w.\u00B7-\uFFFF]/),x=v(/^aria-[\-\w]+$/),S=v(/^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i),L=v(/^(?:\w+script|data):/i),E=v(/[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205f\u3000]/g),M="function"==typeof Symbol&&"symbol"==typeof Symbol.iterator?function(e){return typeof e}:function(e){return e&&"function"==typeof Symbol&&e.constructor===Symbol&&e!==Symbol.prototype?"symbol":typeof e},k=("undefined"!=typeof Reflect&&Reflect).apply,w=Array.prototype.slice,_=Object.freeze,O=function(){return"undefined"==typeof window?null:window};k||(k=function(e,t,n){return e.apply(t,n)});var N=function(e,t){if("object"!==(void 0===e?"undefined":M(e))||"function"!=typeof e.createPolicy)return null;var n=null;t.currentScript&&t.currentScript.hasAttribute("data-tt-policy-suffix")&&(n=t.currentScript.getAttribute("data-tt-policy-suffix"));var r="dompurify"+(n?"#"+n:"");try{return e.createPolicy(r,{createHTML:function(e){return e}})}catch(e){return console.warn("TrustedTypes policy "+r+" could not be created."),null}};return r()});
```
raymondfeng: Is this file copied from somewhere? What's the license?

raymondfeng:
It would be nice to see the commit to be broken down in 3 smaller ones:

- Backend changes
- Seed data
- Front-end changes

packages/shopping/public/purify.js
hacksparrow: [MPL-2.0 OR Apache-2.0](https://github.com/cure53/DOMPurify/blob/master/package.json#L127)

packages/shopping/public/templates.js
hacksparrow: > an easier way to feed the data we got from APIs into the UI element

Do we want to use a framework only for feeding data to the UI? If we use a framework, we would better be using it for every aspect of the app. If we develop the app using a framework, the app will not be as simple as it is now.

I prefer Vue to Angular and React. How do we decide which one to use? Majority vote? I think insights to front-end development and developer experience in general should count more than majority vote. Now, how do we measure that?

Let's keep the app as-is with focus on integrating LoopBack APIs to a framework-neutral app.

Feeding server data to the UI will look more "natural" once we refactor the code to [use async-await](https://github.com/strongloop/loopback4-example-shopping/issues/478).

packages/shopping/package.json
hacksparrow: Errr, thanks for noticing. Fixed it.

hacksparrow:
> It would be nice to see the commit to be broken down in 3 smaller ones.

@raymondfeng I can do that before landing if I get the approvals on this PR. Would like to limit conversations on this PR and make those smaller PRs only for smaller diffs and cleaner history.

The reason why this PR is a combination of many changes is because we wanted a working website as a starting point.

packages/shopping/public/purify.js
bajtos: How are we going to keep this file up-to-date? Please add some developer documentation explaining where is this file coming from and how to update it to a newer version (e.g. to fix security vulnerabilities).

packages/shopping/public/product.js
```diff
@@ -0,0 +1,32 @@
...
+          .replace(/#NAME#/gi, DOMPurify.sanitize(product.name))
+          .replace(
+            /#PRICE#/gi,
+            new Intl.NumberFormat('en-US', {
+              maximumSignificantDigits: 3,
+            }).format(DOMPurify.sanitize(product.price)),
+          )
+          .replace(/#UNFORMATTED-PRICE#/g, DOMPurify.sanitize(product.price))
+          .replace(/#IMAGE#/gi, DOMPurify.sanitize(product.image))
+          .replace(/#DESCRIPTION#/gi, DOMPurify.sanitize(product.description))
```
bajtos: This is super brittle. How can we check that `DOMPurify.sanitize` was called in all places where it's required? How to ensure we don't forget to call `DOMPurify.sanitize` when adding more features in the future?

I think it would be much easier to switch to Handlebars as I mentioned earlier, rather than adding more hacks to the already hacky template rendering. Anyhow 🤷‍♂️

hacksparrow: Let's switch to Handlebars in https://github.com/strongloop/loopback4-example-shopping/issues/478#issuecomment-574176625.

packages/shopping/public/purify.js
hacksparrow: Added https://github.com/strongloop/loopback4-example-shopping/pull/468/commits/46fc4f00b879a50b50080e2754bf0d6a4e7ccdda

raymondfeng: We need to add a `README` or `LICENSE` to have a link to the source and license. Is the dist available on a CDN that we can use?

hacksparrow: Added link to https://github.com/cure53/DOMPurify/blob/master/LICENSE

packages/shopping/src/application.ts
```diff
@@ -127,4 +145,43 @@ export class ShoppingApplication extends BootMixin(
...
+    for (const file of productFiles) {
+      if (file.endsWith('.yml')) {
+        const productFile = path.join(productsDir, file);
+        const yamlString = fs.readFileSync(productFile, 'utf8');
+        const product = YAML.parse(yamlString);
+        await productRepo.create(product);
+      }
+    }
+
+    const passwordHasher = await this.get(
```
jannyHou: you can inject the passwordhasher by `@inject(PasswordHasherBindings.PASSWORD_HASHER)`

hacksparrow: Will do in https://github.com/strongloop/loopback4-example-shopping/issues/478.

hacksparrow: I am not able to reproduce the browser popup. Let's revisit it during #478.

PR #502, opened 2020-01-20
Authorization

hacksparrow:
Implementing authorization based on roles. This is a work in progress, looking for early feedback.


packages/shopping/public/api.js
```diff
@@ -3,6 +3,72 @@
...
+      headers: {
+        Authorization: `Bearer ${token}`,
+      }
+    }).promise();
+  },
+
+  getOrders() {
+    const userId = localStorage.getItem('shoppyUserId');
+    const token = localStorage.getItem('shoppyToken');
+    const roles = localStorage.getItem('shoppyRoles').split(',');
```
dougal83: How do you stop the user altering `shoppyUserId` and `shoppyRoles` in localStorage?  Apologies if I'm jumping the gun on the WIP! 🤐

hacksparrow: The token will be invalid if they edit anything, so request will be rejected.

packages/shopping/src/application.ts
```diff
@@ -192,5 +194,13 @@ export class ShoppingApplication extends BootMixin(
         await userRepo.userCredentials(user.id).create({password});
       }
     }
+
+    // Delete existing shopping carts
+    const cartRepo = await this.getRepository(ShoppingCartRepository);
+    await cartRepo.deleteAll();
+
+    // Delete existing orders
+    const oderRepo = await this.getRepository(OrderRepository);
```
emonddr: > oderRepo 

orderRepo

packages/shopping/public/util.js
```diff
@@ -38,25 +38,39 @@ function parseYml(string) {
...
+        cb && cb(user);
       },
       function(fail) {
-        cb(false);
+        cb && cb(false);
       },
     );
   } else {
-    return cb(false);
+    cb && cb(false);
```
emonddr: does it hang when you specified `return` ? just curious

packages/shopping/public/api.js
```diff
@@ -3,6 +3,33 @@
...
+    }).promise();
+  },
+
+  async updateUser(userId, body) {
+    const token = localStorage.getItem('shoppyToken');
+    const url = apiUrl + `/users/${userId}`;
+    return $.ajax({
+      type: 'POST',
+      url: url,
+      data: JSON.stringify(body),
```
emonddr: why do we need to stringify body here? isn't $.ajax taking in a JSON object anyways? why would obj.data need to be stringified then? Just curious.

emonddr: https://api.jquery.com/jquery.ajax/ says data can be PlainObject or string.

packages/shopping/src/__tests__/acceptance/authorization.acceptance.ts
```diff
@@ -38,9 +31,14 @@ describe('authorization', () => {
     if (app != null) await app.stop();
   });
 
-  describe('customer_service', () => {
-    it('allows customer_service create orders', async () => {
-      newUser = await createAUser();
+  describe('Customer Support', () => {
+    it('does not allows customer support to create orders', async () => {
```
emonddr: > does not allows customer support

does not allow customer support

packages/shopping/src/__tests__/acceptance/authorization.acceptance.ts
```diff
@@ -38,9 +31,14 @@ describe('authorization', () => {
...
+      newUser = await createAUser({
+        email: 'support@loopback.io',
```
emonddr: In earlier commits we removed `email`. So we're putting it back in?

emonddr: Why is it in mocha tests but not web ui or server code?

packages/shopping/public/ui.js
```diff
@@ -232,17 +211,15 @@ async function displayShoppingCart() {
...
-  api.getShoppingCartItems(function(result) {
-    result.items.forEach(item => {
-      if (item.productId === productId) {
-        item.quantity = +quantity;
-      }
-    });
-    api.updateCart(result.items, function() {
-      updatePrice(quantity);
-    });
+async function updateCount(productId, quantity) {
```
emonddr: Why isn't this:

```
updateCount({productId, quantity})
```

like the other ones you change to be ` ( { parm1,...parmN} ) `  ?

hacksparrow: Good suggestion.

packages/shopping/src/__tests__/acceptance/authorization.acceptance.ts
hacksparrow: Email is required for creating the user. The place where I removed is UserProfile which is used from creating the access token, we have no need for email there.

packages/shopping/src/services/basic.authorizor.ts
```diff
@@ -17,30 +18,40 @@ export async function basicAuthorization(
...
   let currentUser: UserProfile;
   if (authorizationCtx.principals.length > 0) {
     const user = _.pick(authorizationCtx.principals[0], [
       'id',
       'name',
-      'email',
+      'roles',
     ]);
-    currentUser = {[securityId]: user.id, name: user.name, email: user.email};
+    currentUser = {[securityId]: user.id, name: user.name, roles: user.roles};
```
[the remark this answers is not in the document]
hacksparrow: These details are made available in the JWT token. We have no need for email.

packages/shopping/src/__tests__/acceptance/authorization.acceptance.ts
emonddr: When up until now, the unique identifier in shopping was a unique email. But a unique userid is fine too. when the rbac csv file was there, bob and alice weren't sufficient for my taste. But now that we have unique user id , then this is fine. we don't need email in the token. thx. @hacksparrow .

packages/shopping/public/api.js
hacksparrow: For some reason it's not working. Let's refactor some other time.

packages/shopping/src/services/basic.authorizor.ts
```diff
@@ -0,0 +1,46 @@
...
+import {UserProfile, securityId} from '@loopback/security';
+
+interface MyAuthorizationMetadata extends AuthorizationMetadata {
+  currentUser?: UserProfile;
+  decision?: AuthorizationDecision;
+}
+
+// Instance level authorizer
+// Can be also registered as an authorizer, depends on users' need.
+export async function basicAuthorization(
```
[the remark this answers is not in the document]
hacksparrow: Basic authorization is an example of an authorization implementation. It implements the following:

1. `@authorize()` any controller method to enforce authorization, else it will be accessible to all.
2. Users with `admin` or `support` roles can access any controller method, IF the role was included in `@authorize({allowedRoles: [...]`.
3. To allow access only to a customer include only `customer` in `allowedRoles`. Now even `admin` and `support` cannot access these methods.

packages/shopping/src/controllers/shopping-cart.controller.ts
```diff
@@ -110,11 +145,16 @@ export class ShoppingCartController {
...
     },
   })
+  @authenticate('jwt')
   async addItem(
     @param.path.string('userId') userId: string,
     @requestBody({description: 'shopping cart item'}) item: ShoppingCartItem,
   ): Promise<ShoppingCart> {
-    debug('Add item %j to shopping cart %s', item, userId);
-    return this.shoppingCartRepository.addItem(userId, item);
+    if (this.currentUserProfile[securityId] === userId) {
```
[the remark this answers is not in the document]
hacksparrow: Indeed! Thanks.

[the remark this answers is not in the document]
hacksparrow: Forgot to update it.

packages/shopping/src/models/order.model.ts
```diff
@@ -15,6 +15,12 @@ export class Order extends Entity {
   })
   orderId?: string;
 
+  @property({
+    type: 'date',
+    id: true,
```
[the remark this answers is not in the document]
hacksparrow: File not included now.

packages/shopping/src/controllers/order.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+        content: {
+          'application/json': {
+            schema: getModelSchemaRef(Order, {includeRelations: true}),
+          },
+        },
+      },
+    },
+  })
+  @authenticate('jwt')
+  async findById(
```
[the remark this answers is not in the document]
hacksparrow: File excluded.

packages/shopping/src/controllers/order.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+  @get('/orders/count', {
+    responses: {
+      '200': {
+        description: 'Order model count',
+        content: {'application/json': {schema: CountSchema}},
+      },
+    },
+  })
+  @authenticate('jwt')
+  async count(
```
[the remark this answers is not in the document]
hacksparrow: File excluded now.

packages/shopping/src/controllers/user-order.controller.ts
```diff
@@ -45,19 +46,12 @@ export class UserOrderController {
     },
   })
   @authenticate('jwt')
-  @authorize({resource: 'order', scopes: ['create']})
+  @authorize({resource: 'userOrder', scopes: ['create'], voters: [basicAuthorization]})
```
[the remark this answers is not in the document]
hacksparrow: File excluded.

packages/shopping/src/controllers/order.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @del('/orders/{id}', {
+    responses: {
+      '204': {
+        description: 'Order DELETE success',
+      },
+    },
+  })
+  @authenticate('jwt')
```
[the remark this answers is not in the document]
hacksparrow: File excluded.

packages/shopping/src/services/basic.authorizor.ts
```diff
@@ -17,30 +18,40 @@ export async function basicAuthorization(
...
     return AuthorizationDecision.ALLOW;
+  }
 
-  const userId = authorizationCtx.invocationContext.args[0];
-  return userId === currentUser[securityId]
-    ? AuthorizationDecision.ALLOW
-    : AuthorizationDecision.DENY;
-*/
+  // Allow access only to model owners
+  if (currentUser[securityId] === authorizationCtx.invocationContext.args[0]) {
```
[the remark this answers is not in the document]
hacksparrow: Just reused an existing code. Maybe @jannyHou knows the details? I haven't looked into in depth yet.

hacksparrow:
@jannyHou 

> are we still using casbin?

Nah. Too complicated to have a scalable implementation for the app.

> which endpoints are secured with @authorize() and what is the rule? 

Whichever we think we should enforce authorization.

Rules:
1. No `@authorize()` === anyone can access. 
2. Add `customer`, `support`,  `admin` to `allowedRoles` to allow access. `customer` will be checked if they are the owner of the model; `support`, and `admin` bypasses this check.
3. `support` and  `admin` won't be able to access methods where they are not explicitly added in `allowedRoles` .

> how many authorizers do we have?

Just a single one. The focus is to show how an authorizer works. We can add more functionality later to show how multilple authorizers can work together.

packages/shopping/src/services/basic.authorizor.ts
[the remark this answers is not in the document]
raymondfeng: The `ctx.args` contains params to be used to invoke the target method. It has nothing to do with http layer.

dougal83:
Nice.  Great job! 👍

packages/shopping/src/services/basic.authorizor.ts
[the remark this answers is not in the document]
InvictusMB: @emonddr I came here looking for your comment. It was absolutely unclear to me how the resource owner is identified. I was looking for something at the model level. I could not imagine that this is related to routing.
That comment line definitely has to be expanded.

I would probably even argue that the last clause should be separated into its own voter with a detailed explanation and a hint on how to implement per resource authorization if the principal ID is not reflected in the route.

PR #505, opened 2020-01-21
feat: enable migrateSchema so that users/products are populated with npm start

raymondfeng:
Before this PR, the `Shoppy` login fails as there are no pre-defined users/products populated.

@hacksparrow Do you have docs to describe that?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,16 +15,15 @@
   },
   "files.insertFinalNewline": true,
   "files.trimTrailingWhitespace": true,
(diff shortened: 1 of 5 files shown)


hacksparrow:
@raymondfeng AppVeyor keeps timing out. Any idea what's up?

raymondfeng:
> @raymondfeng AppVeyor keeps timing out. Any idea what's up?

It may be due to the extra time to seed the database. I improved the PR by adding a flag to control if `databaseSeeding` is needed.

PR #537, opened 2020-02-07
[POC] add refreshtoken to extend accessToken

derdeka:
Proof of concept for https://github.com/strongloop/loopback-next/issues/4573

Early feedback welcome.

Basic Idea:
- Short-lived `accessToken` (e.g. 10 minutes) and very long-lived `refreshToken` (e.g. 48 hours) which are issued at login time
- the `accessToken` giving access to resources
- the `refreshToken` can be used to to create a new `accessToken`
- the `refreshToken` can be revoked, e.g. by logout

### DIFF
--- a/packages/shopping/src/controllers/specs/user-controller.specs.ts
+++ b/packages/shopping/src/controllers/specs/user-controller.specs.ts
@@ -41,3 +41,22 @@ export const CredentialsRequestBody = {
     'application/json': {schema: CredentialsSchema},
   },
 };
+
+const RefreshTokenSchema = {
+  type: 'object',
+  required: ['refreshToken'],
+  properties: {
+    refreshtoken: {
+      type: 'string',
+      minLength: 8,
+    },
+  },
+};
+
+export const RefreshTokenRequestBody = {
+  description: 'The input of refresh function',
+  required: true,
+  content: {
+    'application/json': {schema: RefreshTokenSchema},
+  },
+};
--- a/packages/shopping/src/controllers/user.controller.ts
+++ b/packages/shopping/src/controllers/user.controller.ts
@@ -28,6 +28,7 @@ import {UserProfile, securityId, SecurityBindings} from '@loopback/security';
 import {
   CredentialsRequestBody,
   UserProfileSchema,
+  RefreshTokenRequestBody,
 } from './specs/user-controller.specs';
 import {Credentials} from '../repositories/user.repository';
 import {PasswordHasher} from '../services/hash.password.bcryptjs';
@@ -36,10 +37,12 @@ import {
   TokenServiceBindings,
   PasswordHasherBindings,
   UserServiceBindings,
+  RefreshtokenServiceBindings,
 } from '../keys';
 import _ from 'lodash';
 import {OPERATION_SECURITY_SPEC} from '../utils/security-spec';
 import {basicAuthorization} from '../services/basic.authorizor';
+import {RefreshtokenService} from '../services/refreshtoken.service';
 
 @model()
 export class NewUserRequest extends User {
@@ -59,6 +62,8 @@ export class UserController {
     public passwordHasher: PasswordHasher,
     @inject(TokenServiceBindings.TOKEN_SERVICE)
     public jwtService: TokenService,
+    @inject(RefreshtokenServiceBindings.REFRESHTOKEN_SERVICE)
+    public refreshtokenService: RefreshtokenService<User>,
     @inject(UserServiceBindings.USER_SERVICE)
     public userService: UserService<User, Credentials>,
   ) {}
@@ -250,7 +255,7 @@ export class UserController {
   })
   async login(
     @requestBody(CredentialsRequestBody) credentials: Credentials,
-  ): Promise<{token: string}> {
+  ): Promise<{token: string; refreshtoken: string}> {
     // ensure the user exists, and the password is correct
     const user = await this.userService.verifyCredentials(credentials);
 
@@ -259,6 +264,52 @@ export class UserController {
 
     // create a JSON Web Token based on the user profile
     const token = await this.jwtService.generateToken(userProfile);
+
+    // create a refreshtoken
+    const refreshtoken = await this.refreshtokenService.generateRefreshtoken(
+      user,
+    );
+
+    return {
+      token,
+      refreshtoken,
+    };
+  }
+
+  @post('/users/newtoken', {
+    security: OPERATION_SECURITY_SPEC,
+    responses: {
+      '200': {
+        description: 'Refreshing the token',
+        content: {
+          'application/json': {
+            schema: {
+              type: 'object',
+              properties: {
+                token: {
+                  type: 'string',
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+  })
+  @authenticate('jwt')
+  // TODO(derdeka) find out why @authorize.allowAuthenticated() is not working
+  async refresh(
+    @inject(SecurityBindings.USER) currentUserProfile: UserProfile,
+    @requestBody(RefreshTokenRequestBody) body: {refreshtoken: string},
+  ): Promise<{token: string}> {
+    // check if the provided refreshtoken is valid, throws error if invalid
+    await this.refreshtokenService.verifyRefreshtoken(
+      body.refreshtoken,
+      currentUserProfile,
+    );
+
+    // create a new JSON Web Token based on the user profile
+    const token = await this.jwtService.generateToken(currentUserProfile);
 
     return {token};
   }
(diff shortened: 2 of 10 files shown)


derdeka:
Requesting feedback on this PR. If this is the right direction, i'll add some tests.

packages/shopping/src/models/user.model.ts
```diff
@@ -51,6 +52,9 @@ export class User extends Entity {
   @hasOne(() => UserCredentials)
   userCredentials: UserCredentials;
 
+  @hasMany(() => UserRefreshtoken)
```
dougal83: Why not hasOne? Can user have multiple?

achrinza:
@dougal83 +1 for using an expiration instead of TTL. As per-[RF7519](https://tools.ietf.org/html/rfc7519), the claim should be named `exp`.

packages/shopping/src/controllers/specs/user-controller.specs.ts
```diff
@@ -41,3 +41,22 @@ export const CredentialsRequestBody = {
...
   },
 };
+
+const RefreshTokenSchema = {
+  type: 'object',
+  required: ['refreshToken'],
+  properties: {
+    refreshtoken: {
+      type: 'string',
+      minLength: 8,
```
emonddr: > minLength: 8

Usually the jwt tokens we generate are very long. What is the reasoning for 8? Why not 1? Or why specify this at all? Just curious

packages/shopping/src/services/refreshtoken.service.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
emonddr: @derdeka , cool stuff. I was going to just increase the token lifetime from 10 minutes to 6 hours in my last PR ;).

I am wondering : 
- are you going to add mocha tests?
- have you interacted with the shopping cart API Explorer to see how this will affect the user experience? 
   Developers may want to use the API Explorer to interact with the REST API to debug how things work
   for extended periods of time. It would annoy me when the jwt token would expire in 10 minutes. So I would
   set it for 6 hours while I debug and play with the REST API. How will these changes affect this?
- What needs to change in the Shoppy application (UI front-end) that was recently added?

dougal83: @emonddr All good questions.  Above @derdeka said he'd add tests after initial feedback.

I think the general idea is for the refresh token to hold the session open on a device for a period of time(say 6 hours, two days or even longer) without the need to log back in.  When your access token runs out after 10 minutes, then the refresh token would be used to get new access token without the need to log in again.  This would also need to be designed so that the refresh token can be revoked and force a login. Is that correct @derdeka?

packages/shopping/src/controllers/specs/user-controller.specs.ts
jannyHou: same question

packages/shopping/src/services/refreshtoken.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    userProfile?: UserProfile,
+  ): Promise<UserProfile> {
+    try {
+      if (!userProfile || !userProfile[securityId]) {
+        throw new HttpErrors.Unauthorized('Invalid refreshToken');
+      }
+      const {creation, ttl} = await this.userRefreshtokenRepository.findById(
+        refreshtoken,
+        {
+          where: {
```
jannyHou: usually findById should return only one result, any reason adding an additional where filter here?

packages/shopping/src/controllers/specs/user-controller.specs.ts
jannyHou: @emonddr If you look at the `generateToken` function in
https://github.com/strongloop/loopback4-example-shopping/pull/537/files#diff-90f4ee64422c3d610fc44a185e6d0781R47

A refresh token is the id of a `UserRefreshToken` record, not a token generated by jwt service.

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -260,6 +265,52 @@ export class UserController {
...
+              },
+            },
+          },
+        },
+      },
+    },
+  })
+  @authenticate('jwt')
+  // TODO(derdeka) find out why @authorize.allowAuthenticated() is not working
+  async refresh(
```
jannyHou: When would the refresh function get called? I am trying to illustrate the flow of how a refresh token works: first a user login, then it gets the token and refresh token, then how does it decide when to call the `refresh` endpoint to extend the login time?

packages/shopping/src/models/user.model.ts
derdeka: Currently the `refreshToken` is newly generated at login time. As you can login multiple times or from multiple browsers/devices this needs to be a `hasMany` relation.

Alternatively we could share the `refreshToken` accross browsers/devices, than a `hasOne` would fit. But this might decrease security!?

WDYT?

packages/shopping/src/controllers/specs/user-controller.specs.ts
derdeka: I would prefer a pattern check to the token, but i'm still not sure what is the best way to generate a `refreshToken`. Currently it's a `objectId` which might not be the best practice. On a `mysql`/`postgresql` database i would just use a UUID as `refreshToken` and use a regular expression for validation.

Do you have any other ideas how to generate  a revokeable token?

packages/shopping/src/models/user.model.ts
dougal83: For simplicity, sharing a single token across devices would have a single point of failure, any detected breach would log out all devices. Could be annoying to user.

Your solution is more flexible/complex, but might think about identifying the authorised device linked to the refresh token maybe?  Identifying devices, allows the token to be individually invalidated (i.e. Lost device can have access revoked, although any serious security will remote wipe devices). If one token is compromised, is there any practical difference in access compromised?

Both solutions allow:
- Local - individual devices to be logged out(delete local token)
- Server - all devices to be logged out(expire one token vs. expire all tokens)
- Server - update permissions (update one token vs. update all tokens)

Multiple refresh token adds:
- Server - log out single device(given identity is known, expire the individual token)

Do what you think fits the scenario best.  I'm just offering thoughts for consideration.

packages/shopping/src/services/refreshtoken.service.ts
derdeka: > * are you going to add mocha tests?

I'll add some tests.

> * have you interacted with the shopping cart API Explorer to see how this will affect the user experience?

On a `angular` application i would use a `serviceworker` or `service` to periodically refresh the accesstoken. Is there something similar in the shopping example?

>   Developers may want to use the API Explorer to interact with the REST API to debug how things work
>   for extended periods of time. It would annoy me when the jwt token would expire in 10 minutes. So I would
>   set it for 6 hours while I debug and play with the REST API. How will these changes affect this?

Good catch, i have currently no concept for this. My use case is a regular user on a `angular `SPA.

> * What needs to change in the Shoppy application (UI front-end) that was recently added?

I did not have a deep look how the frontend is currently working. I guess a `serviceworker` could work here too?

packages/shopping/src/services/refreshtoken.service.ts
derdeka: The idea is to check if the `refreshToken` is valid for / belongsTo the current user, validated by the authorization system. It's a kind of additional security check.

packages/shopping/src/controllers/user.controller.ts
derdeka: > then how does it decide when to call the refresh endpoint to extend the login time?

In a `angular` application i would use a `serviceworker` or service with a timer to call this endpoint periodically in background.

raymondfeng:
@derdeka Do see opportunity to add some code to https://github.com/strongloop/loopback-next/pull/4746?

packages/shopping/src/services/refreshtoken.service.ts
derdeka: I recently lerned from @raymondfeng that findById does not respect the where condition. I have to do the check differently.

dhmlau:
@derdeka, the `@loopback/authentication-jwt` extension now has the refresh token capability. Would you be interested to rework this PR to use that instead? Thanks.

cc @jannyhou

mrmodise:
> @derdeka, the `@loopback/authentication-jwt` extension now has the refresh token capability. Would you be interested to rework this PR to use that instead? Thanks.
> 
> cc @jannyHou

Based on this comment I recommend we change the status of the PR to draft?

jannyHou:
@mrmodise I am ok with making the POC PR a draft, or if the gaps are big, creating a new PR is probably better. I would respect @derdeka 's decision on how to processing the feature.

PR #538, opened 2020-02-07
fix: add operation spec to secured endpoints

emonddr:
- Remove unneeded `MyAuthorizationMetadata` interface
- Remove unneeded `compareId` authorizer
- add operation spec to all endpoints with
  @authenticate and @authorize decorators to
  allow usage of these endpoints in the
  API Explorer. (Currently there are many secured endpoints that do not have the operation spec specified, and interacting with these fail because the API explorer doesn't not place the token into the Authorization header)

# Acceptance Criteria
- [x] `npm test` passes 
- [x] All secured endpoints have a `lock` icon in the API Explorer


emonddr:
> @emonddr are you adding the `OPERATION_SECURITY_SPEC` for ALL the endpoints?
> 
> * I don't know if other endpoints like `/products/*` are ready for enabling authentication
> * If all of them are ready & share the same endpoint level scheme, then we should use global scheme instead...

For people that don't use Shoppy, and want to test different endpoints in API explorer. It is frustrating to get the error message that the token is not in the header of the request.

I only added the operation spec to the endpoints currently decorated with 
@authorize and @authenticated.

I am assuming that because they were decorated by @hacksparrow , that they are ready and have been tested.

emonddr:
> @emonddr are you adding the `OPERATION_SECURITY_SPEC` for ALL the endpoints?
> 
> * I don't know if other endpoints like `/products/*` are ready for enabling authentication
> * If all of them are ready & share the same endpoint level scheme, then we should use global scheme instead...

- products.controller.ts : not all endpoints are secured
- shopping-cart.controller.ts : all endpoints are secured
- user-order.controller.ts: all endpoints secured
- user.controller.ts : not all endpoints are secured 

So for some controllers, all endpoints are secured...for others...not all endpoints are secured.

In https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#specifying-the-security-settings-in-the-openapi-specification we talk about setting it at `operation level` or `application level`.

For the latter:

```
import {SECURITY_SCHEME_SPEC, SECURITY_SPEC} from './utils/security-spec';

export class ShoppingApplication extends BootMixin(
  ServiceMixin(RepositoryMixin(RestApplication)),
) {
  constructor(options?: ApplicationConfig) {
    super(options);

    this.api({
      openapi: '3.0.0',
      info: {title: pkg.name, version: pkg.version},
      paths: {},
      components: {securitySchemes: SECURITY_SCHEME_SPEC},
      servers: [{url: '/'}],
      security: SECURITY_SPEC
    });
// ...
```

But this would do it for the entire application; it would secure all endpoints from all controllers, right?

So if a user has a mixture of some controllers with all endpoints secured, and some controllers with only some endpoints secured, then they cannot specify the security spec at the application level...they must specify it all the endpoint level of all controllers. Correct?

@jannyHou  ^

emonddr:
Removed `MyAuthorizationMetadata` as it was never used.

emonddr:
- Removed unneeded `compareId` authorizer since the app is only using `basicAuthorization` authorizer.
- Increased the life of the jwt token from 10 minutes to 6 hours to make it less frustrating when `trying out` the app, and the api explorer, and shoppy etc.

dougal83:
> But this would do it for the entire application; it would secure all endpoints from all controllers, right?

@emonddr I'm not sure about that.  I think it's a general collection based on reading [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-scheme-object).  Like the top level tags property, it looks more descriptive than those in use with operations.

emonddr:
> > But this would do it for the entire application; it would secure all endpoints from all controllers, right?
> 
> @emonddr I'm not sure about that. I think it's a general collection based on reading [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-scheme-object). Like the top level tags property, it looks more descriptive than those in use with operations.

Hey @dougal83 , thanks for your input.  I did try this last year when I wrote this section: https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#specifying-the-security-settings-in-the-openapi-specification .  Setting it at the app level added `lock icons` on all the endpoints in the API explorer.

dougal83:
@emonddr Oh that is interesting. Hmm, it's possible that it is an error to show all items as locked if no endpoint security was specified.  I guess more reading is required on the subject. :)

emonddr:
> @emonddr Oh that is interesting. Hmm, it's possible that it is an error to show all items as locked if no endpoint security was specified. I guess more reading is required on the subject. :)

The API Explorer has an `Authorize` button and dialog, that let's a user set a jwt token. 
Then for every endpoint with a lock icon, the api explorer puts the jwt token in an authorization header when you click on the [Execute] button.

dougal83:
Ah this might explain it, the implementation is fine.

Source: https://swagger.io/specification/#documentStructure

security | [Security Requirement Object] | A declaration of which security mechanisms can be used across the API. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. Individual operations can override this definition.
-- | -- | --

Seems to act as a default for unspecified endpoints.  Does that sound right?

PR #552, opened 2020-02-16
refactor: change duplicate email error

dougal83:
Change user controller email conflict error to 'UnprocessableEntity' as more appropriate.

### DIFF
--- a/packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
+++ b/packages/shopping/src/__tests__/acceptance/user.controller.acceptance.ts
@@ -148,7 +148,7 @@ describe('UserController', () => {
     const res = await client
       .post('/users')
       .send({...userData, password: userPassword})
-      .expect(409);
+      .expect(422);
 
     expect(res.body.error.message).to.equal('Email value is already taken');
   });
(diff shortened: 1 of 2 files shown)


nabdelgadir:
@dougal83 Can you explain why `UnprocessibleEntity` is more appropriate (just for my knowledge)? Looking at the description for [`Conflict`](https://httpstatuses.com/409) it says: 

> The request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request.

Either way, CI is failing because of changing `HttpError` to `ResponseError`, because `ResponseError` doesn't have a `text` property.

dougal83:
Hey @nabdelgadir 
I thought it might be worth revisiting the code. I originally chose Conflict but its come up in conversation on three separate occasions since that some people believe UnprocessibleEntity is more appropriate.  Basically; UnprocessibleEntity is concerned with the current object(new user) having an issue and Conflict argued that must be an _existing_ object being acted on with conflict.

I'm personally happy, to leave it as is.  If you agree it's fine then I'm in! ;)

dougal83:
Happy to leave as is.

nabdelgadir:
Hi @dougal83, thanks for the explanation! I'm good with changing it or keeping it as it, but I think it's worth seeing what @hacksparrow and @jannyHou think about it in this case.

PR #566, opened 2020-02-22
chore: remove dist files from root tsconfig + update node_modules pattern

dougal83:
This speeds up `npm run eslint`, good practice

see https://github.com/strongloop/loopback-next/pull/4707

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,12 +2,7 @@
   "$schema": "http://json.schemastore.org/tsconfig",
   "extends": "@loopback/build/config/tsconfig.common.json",
   "include": ["packages"],
-  "exclude": [
-    "node_modules/**",
-    "packages/*/node_modules/**",
-    "packages/*/dist/**",
-    "**/*.d.ts"
-  ],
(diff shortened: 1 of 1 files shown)


raymondfeng:
We should run the following commands to understand the impact:

1. `rm .eslintcache`
2. `npm run eslint -- --debug`
3. open .eslintcache to see what files are linted
4. change some TS files
5. `npm run eslint --debug` to see how the cache is leveraged to speed up eslint

raymondfeng:
Some observations:

Before I pull in this commit, `dist` does not seem to be linted or cached by `eslint`:

1. The `.eslintcache` file does not have any entries from `dist/**/*.ts`
2. `npm run eslint -- --debug` does not show any mentioning of `dist`

dougal83:
> Some observations:
> 
> Before I pull in this commit, `dist` does not seem to be linted or cached by `eslint`:
> 
> 1. The `.eslintcache` file does not have any entries from `dist/**/*.ts`
> 2. `npm run eslint -- --debug` does not show any mentioning of `dist`

Fair enough, it was unnecessary in that case.  TBH I didn't look deeper than comparing `time npm run eslint` results.  Thank you for checking.

PR #577, opened 2020-02-27
[Spike] angular frontend with openapi-generator sdk

derdeka:
This spike aims to implement an `angular` frontend with the help of [OpenApi Generator](https://github.com/OpenAPITools/openapi-generator).

Early feedback and contributors who like to join the spike welcomed.

Related to: https://github.com/strongloop/loopback-next/issues/4610


packages/angular/src/app/shared/sdk/model/inline-response-200-1.ts
```diff
@@ -0,0 +1,17 @@
...
+ * The version of the OpenAPI document: 1.1.1
+ * 
+ *
+ * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
+ * https://openapi-generator.tech
+ * Do not edit the class manually.
+ */
+
+
+export interface InlineResponse2001 {
```
dougal83: To resolve the responses to meaningful model names you need to ensure that the response schemas have a title property.  The following would resolve `InlineResponse2001` to `AuthResponse`:

```ts
{
  type: 'object',
  title: 'AuthResponse',
  properties: {
    token: {
      type: 'string',
    },
  },
}
```

dougal83: I see you've done it further down! 👍

PR #578, opened 2020-02-27
[Spike] angular frontend with ng-openapi-gen sdk

derdeka:
This spike aims to implement an `angular` frontend with the help of [ng-openapi-gen](https://github.com/cyclosproject/ng-openapi-gen).

Early feedback and contributors who like to join the spike welcomed.

Related to: https://github.com/strongloop/loopback-next/issues/4610


hacksparrow:
👏 excellent!

dougal83:
Great.  What are your thoughts on hosting the angular app via express alongside the loopback application?  How come you chose this particular generator did you look at others?

derdeka:
> What are your thoughts on hosting the angular app via express alongside the loopback application?

I normally prefer to have seperate projects or even seperate repositories for frontend and backend as the builtin angular tooling is complex, changes frequently and built takes a while. In my projects i just deploy the generated static bundle files of the progressive webapp to `aws s3` by continous deployment `aws codepipeline`. No (express) server needed at all.

> How come you chose this particular generator did you look at others?

I did some POCs with several sdks, but none of them fits my needs. I'm just uploading here parts of my researches, ask for community feedback and still hope to find a better sdk.
If you have any other idea for an sdk we should test, please add your suggestions to the main thread https://github.com/strongloop/loopback-next/issues/4610

dougal83:
It would be great to have a number of deployment options demonstrated as a stretch goal.  It would make loopback/angular combo very accessible.

Personally I do like the generator in the [other spike](https://github.com/strongloop/loopback4-example-shopping/pull/577) as it opens the door to many other options apart from angular.

Thanks for sharing these, they are great examples.

dhmlau:
@derdeka, thanks for looking into this! 

I have two comments/questions:
- If just comparing the angular generator, what are the differences between `ng-openapi-gen` and `openapi-generator` (you did it for #577)? Any pros and cons of using one over the other?
- I think this spike can serve two purposes: 
    1. If I create a LB4 app from scratch and now want to build a frontend, is there any generator we would recommend?  We may not need to choose it for users but if one is better (in terms of ease of use, more comprehensive client, etc), we do make certain recommendation. And if we want to build an e2e tutorial, it helps us decide which one to pick. 
   2. If I have a LB3 app and generated client using the Angular SDK, which path would make my migration process easier?  I'm looking at the generated client in the [loopback-example-angular](https://github.com/strongloop/loopback-example-angular/tree/master/client/js), does it mean the `*.controller.service.ts` generated in this PR would map to the [controllers in the /controllers folder](https://github.com/strongloop/loopback-example-angular/tree/master/client/js/controllers)?

dhmlau:
FYI - I added this docs page to contain steps to create the model and web client in an angular application using `ng-openapi-gen`. See https://github.com/strongloop/loopback-next/blob/master/docs/site/Building-frontend-angular-application.md

PR #582, opened 2020-03-01
fix: remove ref for v4.loopback.io

dhmlau:
Since v4.loopback.io is no longer in use, remove any references for v4.loopback.io.


dhmlau:
@slnode test please

PR #610, opened 2020-03-22
refactor: update auth strategy binding

dougal83:
utilise new binding method and include securityScheme

### DIFF
--- a/packages/shopping/src/application.ts
+++ b/packages/shopping/src/application.ts
@@ -3,48 +3,48 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {AuthenticationComponent} from '@loopback/authentication';
+import {AuthorizationComponent} from '@loopback/authorization';
 import {BootMixin} from '@loopback/boot';
-import {ApplicationConfig, BindingKey} from '@loopback/core';
 import {
-  RepositoryMixin,
-  SchemaMigrationOptions,
+  ApplicationConfig,
+  BindingKey,
+  createBindingFromClass,
+} from '@loopback/core';
+import {
   model,
   property,
+  RepositoryMixin,
+  SchemaMigrationOptions,
 } from '@loopback/repository';
 import {RestApplication} from '@loopback/rest';
-import {ServiceMixin} from '@loopback/service-proxy';
-import {MyAuthenticationSequence} from './sequence';
 import {
   RestExplorerBindings,
   RestExplorerComponent,
 } from '@loopback/rest-explorer';
-import {
-  TokenServiceBindings,
-  UserServiceBindings,
-  TokenServiceConstants,
-  PasswordHasherBindings,
-} from './keys';
-import {JWTService} from './services/jwt-service';
-import {MyUserService} from './services/user-service';
+import {ServiceMixin} from '@loopback/service-proxy';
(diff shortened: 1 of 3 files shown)


dougal83:
Also, imports sorted by plugin adding noise to changes.  Would be good to have all imports automatically sorted generally.

dougal83:
Question: Should imports from `openapi-v3` package come from the package itself or via `rest`?  

@jannyHou What is the reason for the export below?  Is this for the purpose of importing via `rest`?

https://github.com/strongloop/loopback-next/blob/38fd41e26cd5abf52e69d0b275a730b195be813a/packages/rest/src/index.ts#L6

packages/shopping/src/application.ts
```diff
@@ -93,7 +92,7 @@ export class ShoppingApplication extends BootMixin(
     this.component(AuthorizationComponent);
 
     // authentication
-    registerAuthenticationStrategy(this, JWTAuthenticationStrategy);
+    this.add(createBindingFromClass(JWTAuthenticationStrategy));
```
raymondfeng: We should `registerAuthenticationStrategy` if it does not use `createBindingFromClass`.

dougal83: Since it is application level, should the user elect to use the appropriate binding method based on their auth strategy implementation?

emonddr: > We should `registerAuthenticationStrategy` if it does not use `createBindingFromClass`.

Why are we changing this? The comment is missing some words I think...it makes it unclear why we're making this change.

Also, this approach is documented in : https://loopback.io/doc/en/lb4/Loopback-component-authentication.html#registering-a-custom-authentication-strategy .

dougal83: `this.add(createBindingFromClass(JWTAuthenticationStrategy))` can easily be moved into `registerAuthenticationStrategy` for consistency if that is what is desirable. Say the word.

raymondfeng: @dougal83 The implementation of `registerAuthenticationStrategy` already uses `createBindingFromClass`. What issue did you run into that led to your change?

raymondfeng:
>Question: Should imports from openapi-v3 package come from the package itself or via rest?

Either way is fine if your package has explicit dependencies on both packages. Otherwise, use the one from the `dependencies`.

> @jannyHou What is the reason for the export below? Is this for the purpose of importing via rest?

> https://github.com/strongloop/loopback-next/blob/38fd41e26cd5abf52e69d0b275a730b195be813a/packages/rest/src/index.ts#L6

The exports from `openapi-v3` are re-exported in `rest` to simplify deps. It's also true that `core` reexports `context`.

packages/shopping/src/application.ts
dougal83: @raymondfeng It was just an arbitrary decision on my part.  Since the strategies were app level, I just shortcut to `this.add(createBindingFromClass(JWTAuthenticationStrategy))`.

I'll alter the code to keep `registerAuthenticationStrategy`.

dougal83: @emonddr Please alter the binding as you see fit going forward.  The documented `registerAuthenticationStrategy` is unchanged and I've shortcut to `this.add(createBindingFromClass(JWTAuthenticationStrategy))` as is in this instance.

PR #611, opened 2020-03-22
feat: add info spec enhancer

dougal83:
Initial implementation as a starting point for discussion.

Would we be looking for a core enhancer or a per-app implementation?  Ready for feedback for improvement.

See also: https://github.com/strongloop/loopback-next/pull/3884

### DIFF
--- a/packages/shopping/src/spec-enhancers/info-spec.extension.ts
+++ b/packages/shopping/src/spec-enhancers/info-spec.extension.ts
@@ -1,0 +1,37 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {bind, Context, inject} from '@loopback/context';
+import {
+  asSpecEnhancer,
+  mergeOpenAPISpec,
+  OASEnhancer,
+  OpenApiSpec,
+} from '@loopback/openapi-v3';
+import {PackageInfo, PackageKey} from '../application';
+
+@bind(asSpecEnhancer)
+export class InfoSpecEnhancer implements OASEnhancer {
+  name = 'info';
+  pkg: PackageInfo;
+
+  constructor(@inject.context() private ctx: Context) {
+    this.pkg = ctx.getSync<PackageInfo>(PackageKey);
+  }
+
+  modifySpec(spec: OpenApiSpec): OpenApiSpec {
+    const patchSpec = {
+      openapi: '3.0.0',
+      info: {
+        title: this.pkg.name,
+        version: this.pkg.version,
+        contact: {},
+      },
+      servers: [{url: '/'}],
+    };
+
+    return mergeOpenAPISpec(spec, patchSpec);
+  }
+}
--- a/packages/shopping/src/application.ts
+++ b/packages/shopping/src/application.ts
@@ -3,48 +3,52 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {BootMixin} from '@loopback/boot';
-import {ApplicationConfig, BindingKey} from '@loopback/core';
-import {
-  RepositoryMixin,
-  SchemaMigrationOptions,
-  model,
-  property,
-} from '@loopback/repository';
-import {RestApplication} from '@loopback/rest';
-import {ServiceMixin} from '@loopback/service-proxy';
-import {MyAuthenticationSequence} from './sequence';
-import {
-  RestExplorerBindings,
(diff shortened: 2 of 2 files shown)


dougal83:
See also: https://github.com/strongloop/loopback4-example-shopping/pull/610#issuecomment-602220128

packages/shopping/src/spec-enhancers/info-spec.extension.ts
```diff
@@ -0,0 +1,37 @@
...
+} from '@loopback/openapi-v3';
+import {PackageInfo, PackageKey} from '../application';
+
+@bind(asSpecEnhancer)
+export class InfoSpecEnhancer implements OASEnhancer {
+  name = 'info';
+  pkg: PackageInfo;
+
+  constructor(@inject.context() private ctx: Context) {
+    this.pkg = ctx.getSync<PackageInfo>(PackageKey);
```
raymondfeng: Why not `@inject(PakageKey) private pkg: PackageIno`?

packages/shopping/src/spec-enhancers/info-spec.extension.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  modifySpec(spec: OpenApiSpec): OpenApiSpec {
+    const patchSpec = {
+      openapi: '3.0.0',
+      info: {
+        title: this.pkg.name,
+        version: this.pkg.version,
+        contact: {},
```
raymondfeng: We can set up `info` as part of `@loopback/rest` with these default values.

packages/shopping/src/spec-enhancers/info-spec.extension.ts
dougal83: It complained at me when I tried that so I took path of last resistance and just rewrote. I'll redo and see if it was typescript engine being temperamental.

dougal83: @raymondfeng The error as follows:

```node
UnhandledPromiseRejectionWarning: Error: Cannot resolve injected arguments for InfoSpecEnhancer.[0]: The arguments[0] is not decorated for dependency injection, but a value is not supplied
[0]     at /home/dev/workspace/loopback4-example-shopping/packages/shopping/node_modules/@loopback/context/dist/resolver.js:167:23
```

packages/shopping/src/spec-enhancers/info-spec.extension.ts
dougal83: Like the [core enhancer](https://github.com/strongloop/loopback-next/pull/4365/files#diff-576584d72083bead0929c10dfc101b01R256) proposal

packages/shopping/src/spec-enhancers/info-spec.extension.ts
raymondfeng: It's weird. I'll troubleshoot.

packages/shopping/src/spec-enhancers/info-spec.extension.ts
raymondfeng: Either that, or just hard-code such defaults in `rest-server`.

packages/shopping/src/spec-enhancers/info-spec.extension.ts
```diff
@@ -0,0 +1,37 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {bind, Context, inject} from '@loopback/context';
+import {
+  asSpecEnhancer,
+  mergeOpenAPISpec,
+  OASEnhancer,
+  OpenApiSpec,
+} from '@loopback/openapi-v3';
+import {PackageInfo, PackageKey} from '../application';
```
raymondfeng: Due to the order of `require`, `PackageKey` is `undefined` when this file is loaded. As a result, `@inject` fails because the binding key is `undefined`.

Moving `PackgeKey` from `application.ts` to `keys.ts` fixes the issue.

packages/shopping/src/spec-enhancers/info-spec.extension.ts
raymondfeng: See https://github.com/strongloop/loopback-next/pull/4955

packages/shopping/src/spec-enhancers/info-spec.extension.ts
raymondfeng: See https://github.com/strongloop/loopback4-example-shopping/pull/611/files#r396672461

raymondfeng:
@dougal83 Should we close this PR to favor https://github.com/strongloop/loopback-next/pull/4955?

dougal83:
@raymondfeng Sure. No problemo!

PR #619, opened 2020-03-26
feat: utilise new rest info oasEnhancer

dougal83:
add placeholder author in package.json & remove temp workaround

### DIFF
--- a/packages/shopping/package.json
+++ b/packages/shopping/package.json
@@ -30,6 +30,11 @@
   "repository": {
     "type": "git"
   },
+  "author": {
+    "name": "LoopBack contributors",
+    "email": "dev@loopback.io",
+    "url": "https://loopback.io"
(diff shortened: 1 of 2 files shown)


packages/shopping/package.json
```diff
@@ -30,6 +30,7 @@
   "repository": {
     "type": "git"
   },
+  "author": "Joe Bloggs <joe.bloggs@loopback.io> (https://loopback.io/)",
```
raymondfeng: Maybe use something like `LoopBack Contributors`?

raymondfeng: To set a good example, maybe it's even better to use an object, such as:
```json
"author": {
  "name": "LoopBack contributors",
  "email": "dev@loopback.io",
  "url": "https://loopback.io"
}
```

Issue #17, opened 2018-09-24
Implement skipped tests


virkt25:
There are 3 tests in `user-order.contoller.acceptance.ts` that are skipped because the underlying features needed haven't yet been implemented. 

The missing features are: 
- https://github.com/strongloop/loopback-next/issues/1718
- https://github.com/strongloop/loopback-next/issues/100

Once the above have been implemented, the skipped tests should be implemented. 

---

- [ ] Implement test titled: `throws an error when creating an order for a non-existent user`
- [ ] Implement test titled: `patches orders matching filter for a given user`
- [ ] Implement test titled: `deletes orders matching filter for a given user`

bajtos:
FWIW, `strongloop/loopback-next#100` is done (but not published to npm yet).

PR #622, opened 2020-03-29
feat: Add missing filter tests

dougal83:
Changed params to object and implemented two filter tests from #17


dougal83:
Note noise on imports due to import sorting.

PR #692, opened 2020-05-14
fix: refactor auth tutorial

dhmlau:
I'd like to do some rework in the original [JWT authentication tutorial](https://loopback.io/doc/en/lb4/Authentication-Tutorial.html) which is based on the shopping example. Since it has lots of great content in it, I'd like to preserve it in this repo. 

This PR copies the tutorial https://loopback.io/doc/en/lb4/Authentication-Tutorial.html. The only change is to fix the links since the location has changed. 

cc @jannyHou @deepakrkris

### DIFF
--- a/authentication-tutorial.md
+++ b/authentication-tutorial.md
@@ -1,0 +1,1314 @@
+## Overview
+
+LoopBack 4 has an authentication package `@loopback/authentication` which allows
+you to secure your application's API endpoints with custom authentication
+strategies and an `@authenticate` decorator.
+
+This tutorial showcases how `authentication` was added to the
+[loopback4-example-shopping](https://github.com/strongloop/loopback4-example-shopping)
+application by **creating** and **registering** a custom authentication strategy
+based on the `JSON Web Token (JWT)` approach.
+
+Here is a brief summary of the `JSON Web Token (JWT)` approach.
+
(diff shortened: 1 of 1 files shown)


authentication-tutorial.md
```diff
@@ -0,0 +1,1314 @@
...
+  constructor(options?: ApplicationConfig) {
+    super(options);
+
+    this.api({
+      openapi: '3.0.0',
+      info: {title: pkg.name, version: pkg.version},
+      paths: {},
+      components: {securitySchemes: SECURITY_SCHEME_SPEC},
+      servers: [{url: '/'}],
+    });
```
dougal83, line 1252: You may be able to simplify this part.  (see https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/application.ts).  In built enhancer in place now. 👍

dhmlau:
@dougal83, the content is actually written by @emonddr. :) You reminded me that I should add @emonddr's authorship in this commit. I'm in the middle of simplifying the JWT auth tutorial but the original content, as you said, contains lots of useful information, so I'd like to preserve it in here and see what we can put back in other parts of the documentation. 

See PR https://github.com/strongloop/loopback-next/pull/5421.

dhmlau:
We've done some work in improving the authentication documentation. After discussing with @jannyHou, we think this PR is good to close.

PR #747, opened 2020-06-13
feat: add email check to verifyCredentials

dougal83:
Guard against undefined email that would incorrectly return a user.


hacksparrow:
Thanks @dougal83, merged.

Issue #765, opened 2020-07-02
Fix loopback4-example-recommender not found on npm


hacksparrow:
Because of the way `loopback4-example-recommender` is configured to be used, we are starting to see errors like this.

```
...
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/loopback4-example-recommender - Not found
npm ERR! 404 
npm ERR! 404  'loopback4-example-recommender@1.1.1' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in:
npm ERR!     /tmp/renovate-cache/others/npm/_logs/2020-06-25T12_16_21_576Z-debug.log

lerna ERR! npm install --ignore-scripts --ignore-scripts --no-audit --package-lock-only exited 1 in 'loopback4-example-shopping'
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
```

## Acceptance Criteria

- [ ] Find the cause
- [ ] Fix the problem

jannyHou:
@hacksparrow is this issue still valid? the example seems working fine on my local.

hacksparrow:
Caused by some config change, Raymond fixed it.

PR #766, opened 2020-07-02
feat: install recommender as a local package

hacksparrow:
Reconfigured `loopback4-example-recommender` to be installed as a local dependency. Fixes https://github.com/strongloop/loopback4-example-shopping/issues/765 and also removes confusion for developers looking at the repository.

`loopback4-example-recommender` was configured like it was something from the npm registry, while it is not. Due to something (node, npm, ts; not sure what) changing, the package was looked up on npm, where it was not found (expected behavior IMO). I have updated `loopback4-example-recommender` to be installed as a local package and resolved as one.

### DIFF
--- a/packages/shopping/package.json
+++ b/packages/shopping/package.json
@@ -11,6 +11,7 @@
     "node": ">=8"
   },
   "scripts": {
+    "preinstall": "npm i ../recommender/",
     "build": "lb-tsc",
     "build:watch": "lb-tsc --watch",
     "clean": "lb-clean dist* *.tsbuildinfo",
@@ -81,7 +82,7 @@
     "@types/mocha": "7.0.2",
     "@types/node": "12.12.47",
     "concurrently": "5.2.0",
-    "loopback4-example-recommender": "1.1.1",
+    "loopback4-example-recommender": "file:../recommender",
     "mocha": "8.0.1",
     "source-map-support": "0.5.19",
     "typescript": "3.9.5"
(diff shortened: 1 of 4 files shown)


packages/shopping/package.json
```diff
@@ -11,6 +11,7 @@
     "node": ">=8"
   },
   "scripts": {
+    "preinstall": "npm i ../recommender/",
```
raymondfeng, line 14: Please remove it. `lerna bootstrap` will handle it.

packages/shopping/package.json
```diff
@@ -81,7 +82,7 @@
     "@types/mocha": "7.0.2",
     "@types/node": "12.12.47",
     "concurrently": "5.2.0",
-    "loopback4-example-recommender": "1.1.1",
+    "loopback4-example-recommender": "file:../recommender",
```
raymondfeng, line 85: IIRC, `lerna` allows both style.

hacksparrow: Doesn't seem to be working anymore. Take a look here - https://github.com/strongloop/loopback4-example-shopping/pull/760.

raymondfeng:
-1 to change how our lerna monorepo is setup for local dependencies. Let's get to the root of `renovate` issue first.

raymondfeng:
I checked `If you want to rebase/retry this PR, check this box` in failing PRs created by `renovate` to force a retry. The refreshed PRs seem to be good now.

PR #785, opened 2020-07-15
feat: update jwt component

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/5294

- Apply `@loopback/authentication-jwt` to this example.
- Token service and User service are still kept. 
- Custom spec enhancer is added to avoid globally added security policy in @loopback/authentication-jwt.

The imports are automatically fixed by the lint command.


packages/shopping/src/application.ts
```diff
@@ -73,14 +72,12 @@ export class ShoppingApplication extends BootMixin(
...
-    this.setUpBindings();
-
     // Bind authentication component related elements
     this.component(AuthenticationComponent);
+    this.component(JWTAuthenticationComponent);
     this.component(AuthorizationComponent);
 
-    // authentication
-    this.add(createBindingFromClass(JWTAuthenticationStrategy));
+    this.setUpBindings();
```
hacksparrow, line 80: Why do we need this?

hacksparrow: I see it is a method below.

jannyHou: yeah we organize all bindings inside that function.

PR #869, opened 2020-09-21
chore: update to latest LoopBack versions

raymondfeng:
Signed-off-by: Raymond Feng <enjoyjava@gmail.com>

FYI - the default port for grpc 50051 is used by another process on my Mac - /Library/LaunchDaemons/com.canonical.multipassd.plist (The Cannon printer). I have to switch it to `50000` or `0`.


.renovaterc.json
```diff
@@ -32,7 +32,7 @@
   "semanticCommitType": "chore",
   "semanticCommitScope": null,
 
-  "separateMajorMinor": true,
+  "separateMajorMinor": false,
```
hacksparrow, line 35: What's the implication of this change?

packages/recommender/src/recommendation-grpc.ts
```diff
@@ -42,15 +41,15 @@ export function loadRecommendationService() {
  * it serves.
  * @return The new server object
  */
-export function createGRPCRecommendationServer(port = '0.0.0.0:50051') {
+export function createGRPCRecommendationServer(port = '0.0.0.0:0') {
```
hacksparrow, line 44: Below it says `50000`, here it is `0`. Why the mismatch?

raymondfeng: We have two cases:

1. Set `0` to get an available port and use the assigned port for testing
2. Set `50000` as a fixed port so that it can be configured statically in datasource json.

hacksparrow: Can we use `host` instead of `port` because `host` is conventionally `hostname[:port]`. We should update in all other relevant places.

.renovaterc.json
raymondfeng: The current settings creates two PRs for a new LB release: one for major versions and one for other changes. As a result, each PR gets half of the newly released LB packages and both of them fails in many cases.

packages/recommender/src/recommendation-grpc.ts
raymondfeng: That's the grpc term.

hacksparrow (approved):

PR #881, opened 2020-10-06
feat: add password reset functionality

### LINKED ISSUE #818: Add password reset functionality
Password reset functionality is a common requirement, let's add it in the example app. Ref: https://github.com/strongloop/loopback4-example-shopping/issues/456.

mrmodise:
Fixes #818

### DIFF
--- a/packages/shopping/public/ui.js
+++ b/packages/shopping/public/ui.js
@@ -109,9 +109,80 @@ async function logIn(email, password) {
     await refreshLogInStatus();
     $('#logInModal').modal('hide');
     $('#signUpModal').modal('hide');
+    $('#passwordResetModal').modal('hide');
   } else {
     $('#logInTitle').text('Invalid credentials');
   }
+}
+
+async function resetPassword(email, password, confirmPassword) {
+  email = email || $('#passwordResetEmail').val();
+  password = password || $('#passwordResetPassword').val();
+  confirmPassword = confirmPassword || $('#passwordResetConfirmPassword').val();
+
+  if (typeof password === 'undefined' || password.length === 0) {
+    $('#passwordResetNoPassword').show();
+    setTimeout(() => {
+      $('#passwordResetNoPassword').hide();
+    }, 3000);
+    return;
+  }
+
+  if (password.length < 8) {
+    $('#passwordResetNoCorrectPassword').show();
+    setTimeout(() => {
+      $('#passwordResetNoCorrectPassword').hide();
+    }, 3000);
+    return;
+  }
+
+  if (typeof confirmPassword === 'undefined' || confirmPassword.length === 0) {
+    $('#passwordResetNoConfirmPassword').show();
+    setTimeout(() => {
+      $('#passwordResetNoConfirmPassword').hide();
+    }, 3000);
+    return;
+  }
+
+  if (password !== confirmPassword) {
+    $('#passwordResetMismatch').show();
+
+    setTimeout(() => {
+      $('#passwordResetMismatch').hide();
+    }, 3000);
+    return;
+  }
+
+  const res = await api.passwordReset({email, password}).catch(() => {
+    $('#passwordResetFailed').show();
+    setTimeout(() => {
+      $('#passwordResetFailed').hide();
+    }, 3000);
+  });
+
+  if (res) {
+    localStorage.setItem('shoppyToken', res.token);
+    $('#passwordResetChanged').show();
+    setTimeout(() => {
+      $('#passwordResetChanged').hide();
+    }, 3000);
+
+    $('#passwordResetEmail').val('');
+    $('#passwordResetPassword').val('');
+    $('#passwordResetConfirmPassword').val('');
+  }
+}
+
+function displayPasswordReset() {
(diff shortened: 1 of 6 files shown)


dougal83:
Also looking at logs I think you need to fetch the upstream and rebase as you are 75 commits behind the master.

mrmodise:
> Also looking at logs I think you need to fetch the upstream and rebase as you are 75 commits behind the master.

should be good now. Thanks

nflaig (commented):
@mrmodise the current implementation would let anyone change the password of a user account by just providing the email address. There are multiple approaches to implementing password reset but most of the time it involves sending an email to the user. You can checkout this [stackoverflow discussions](https://stackoverflow.com/questions/1102781/best-way-for-a-forgot-password-implementation) to see some discussed approaches.

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -260,4 +261,46 @@ export class UserController {
...
+    // encrypt the new password
+    const passwordHash = await this.passwordHasher.hashPassword(password);
+
+    // reset the password
+    await this.userRepository
+      .userCredentials(user.id)
+      .patch({password: passwordHash});
+
+    // return the user
+    return this.userRepository.findById(user.id);
```
nflaig, line 304: @mrmodise seems unnecessary to do another database operation here since you already have the user model

mrmodise: Agreed and I highlighted this in the comments. I can do the full blown solution but I thought it would be an overkill for a POC

mrmodise:
> @mrmodise the current implementation would let anyone change the password of a user account by just providing the email address. There are multiple approaches to implementing password reset but most of the time it involves sending an email to the user. You can checkout this [stackoverflow discussions](https://stackoverflow.com/questions/1102781/best-way-for-a-forgot-password-implementation) to see some discussed approaches.

Agreed and I highlighted this in the comments. I can do the full blown solution but I thought it would be an overkill for a POC

packages/shopping/src/controllers/user.controller.ts
mrmodise: Updated accordingly

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -260,4 +261,46 @@ export class UserController {
...
+        description: 'The updated user profile',
+        content: {
+          'application/json': {
+            schema: UserProfileSchema,
+          },
+        },
+      },
+    },
+  })
+  async passwordReset(
```
jannyHou, line 277: To make sure the user is updating it's own password, you can inject the CURRENT_USER like [this](https://github.com/strongloop/loopback4-example-shopping/blob/cf7678baaa5dcf64c7ff93728f32f545064cb6c2/packages/shopping/src/controllers/user.controller.ts#L142), then compare the user email.

jannyHou (commented):
@mrmodise Appreciate your help! As the discussion mentioned, we can add a few code to verify the user can only resetting its own password. I commented in the controller file. 

Otherwise the PR looks reasonable to me and test 💯 .

packages/shopping/src/controllers/user.controller.ts
mrmodise: Noted will update accordingly

packages/shopping/src/controllers/user.controller.ts
```diff
@@ -260,4 +261,47 @@ export class UserController {
...
+
+    validateCredentials(_.pick(credentials, ['email', 'password']));
+
+    const passwordHash = await this.passwordHasher.hashPassword(password);
+
+    await this.userRepository
+      .userCredentials(user.id)
+      .patch({password: passwordHash});
+
+    return user;
```
hacksparrow, line 305: Instead of returning the user, please generate a new token and return it, basically a re-login.

mrmodise: Change included

packages/shopping/public/ui.js
```diff
@@ -109,11 +109,82 @@ async function logIn(email, password) {
...
 }
 
+async function resetPassword(email, password, confirmPassword) {
+  email = email || $('#passwordResetEmail').val();
+  password = password || $('#passwordResetPassword').val();
+  confirmPassword = confirmPassword || $('#passwordResetConfirmPassword').val();
+
+  if (typeof password === 'undefined' || password.length === 0) {
+    $('#passwordResetNoPassword').show();
+    setInterval(() => {
```
hacksparrow, line 125: Shouldn't this be `setTimeout`?

hacksparrow: And all the other instances, of course.

hacksparrow:
@mrmodise good to land once `setInterval` is replaced by `setTimeout`.

packages/shopping/public/ui.js
mrmodise: Updated accordingly

hacksparrow (approved):

hacksparrow:
@mrmodise Thank you Mo! Landed.

PR #886, opened 2020-10-10
feat: add cypress automated UI testing framework

mrmodise:
Closes #476

Issue #476, opened 2020-01-13
Shopping App Website: UI testing


hacksparrow:
Automated UI testing (https://webdriver.io/ or https://www.cypress.io).

**Acceptance Criteria**

All aspects of the UI interaction and behavior should be covered by a UI testing framework.

mrmodise:
@hacksparrow am interested in working on this. Please advise on next steps.

hacksparrow:
@mrmodise basically, we will have to ensure the UI - its appearance and interaction aspect - are working as expected and are not broken by accidental changes.

For eg: if the text was changed to "Lg in" from "Log In", it should be detected; clicking the "Log In" button should log in the user (since the form in pre-filled with working creds); trying to log in with wrong creds should not show the logged in UI, etc.

https://www.cypress.io/ or https://webdriver.io/ should help with it.

mrmodise:
Thanks @hacksparrow I have used both and I would recommend Cypress. I can get started right away if you cool with it

hacksparrow:
@mrmodise that would be great.

mrmodise:
> @mrmodise that would be great.

Cool will send something your way ASAP

dougal83:
+1 for Cypress.  Thank you @mrmodise

mrmodise:
> +1 for Cypress. Thank you @mrmodise

@dougal83 they're coming up really well. Branch: https://github.com/mrmodise/loopback4-example-shopping/tree/feature/automated-ui-tests/packages/shopping/cypress/integration

PR #904, opened 2020-10-24
feat: update cypress e2e tests to typescript

mrmodise:
Follow up to #476 based on https://github.com/strongloop/loopback4-example-shopping/pull/886#issuecomment-712296492


dougal83 (approved):
Nice.  All checks out! :+1: 

FYI: One test is a bit inconsistent, failed on first run.
**Profile: should update user profile**
`Timed out retrying: Expected to find content: 'LoopBack Doe' within the element: <div#user.nav-item> but never did.`

mrmodise:
> Nice. All checks out! 👍
> 
> FYI: One test is a bit inconsistent, failed on first run.
> **Profile: should update user profile**
> `Timed out retrying: Expected to find content: 'LoopBack Doe' within the element: <div#user.nav-item> but never did.`

Thanks am going to add `.should('be.visible')` this should resolve it. The fun of using pop ups 😄

dougal83:
> Thanks am going to add `.should('be.visible')` this should resolve it. The fun of using pop ups 😄

I think that would do it but how about a value oriented check?  Would make the test more self descriptive. What do you think?https://docs.cypress.io/guides/references/assertions.html#Value

mrmodise:
> https://docs.cypress.io/guides/references/assertions.html#Value

Doesn't work. Returns the error immediately:

`expected <div#user.nav-item> to have value LoopBack Doe, but the value was ''`

Interesting that it times out:

`Timed out retrying: expected '<div#user.nav-item>' to have value 'LoopBack Doe', but the value was ''`

dougal83:
> > https://docs.cypress.io/guides/references/assertions.html#Value
> 
> Doesn't work. Returns the error immediately:
> 
> `expected <div#user.nav-item> to have value LoopBack Doe, but the value was ''`
> 
> Interesting that it times out:
> 
> `Timed out retrying: expected '<div#user.nav-item>' to have value 'LoopBack Doe', but the value was ''`

I'm guessing that something like would work: `.then(() => cy.get('[data-cy=user]').should('contain', 'LoopBack Doe')),` Sorry, I didn't look in detail at markup before comment.

packages/shopping/src/__tests__/e2e/integration/profile.spec.ts
```diff
@@ -13,17 +13,24 @@ context('Profile', () => {
...
       .type('LoopBack')
       .then(() =>
         cy
           .get('[data-cy=update-profile]')
           .click()
-          .then(() => cy.get('[data-cy=user]').contains('LoopBack Doe')),
+          .then(() =>
+            cy
+              .get('[data-cy=user]')
+              .should('be.visible')
```
dougal83, line 31:
```suggestion
              .should('contain', 'Loopback Doe'),
```

packages/shopping/src/__tests__/e2e/integration/profile.spec.ts
```diff
@@ -13,17 +13,24 @@ context('Profile', () => {
...
+              .contains('LoopBack Doe'),
```
dougal83, line 32:
```suggestion
              
```

packages/shopping/src/__tests__/e2e/integration/profile.spec.ts
mrmodise: Done

nabdelgadir (approved):
Changes LGTM 👍 The profile test is passing for me now as well.

nabdelgadir:
The CI failure is unrelated to your changes. Can you push again so it restarts? Thanks!

dhmlau:
Thanks @mrmodise. You're granted with write access to this repo. So feel free to land PRs in this repo as long as it has approvals. Thanks!

mrmodise:
> Thanks @mrmodise. You're granted with write access to this repo. So feel free to land PRs in this repo as long as it has approvals. Thanks!

Thanks didn't want to seem am approving my own work 🤣

PR #929, opened 2020-11-06
feat: allow using custom JWT secret token

### LINKED ISSUE #925: Make it possible to select JWT secret via environment variable
I think it would be useful if the loopback4-example-shopping allowed setting the JWT secret via an environment variable.
Today the shopping API/application uses the default secret defined in @loopback/authentication-jwt.
I suggest adding code like this to setUpBindings() in packages/shopping/src/application.ts:
```
    let token = process.env.JWTSECRET;
    if (!token) {
      console.error('JWTSECRET not set, using default secret for JWT generation/validation');
    } else {
      this.bind(TokenServiceBindings.TOKEN_SECRET).to(token);
    }    
```
This code serves both as an example of how to use your own secret with @loopback/authentication-jwt and makes it possible to use loopback4-example-shopping in demos together with other JWT generators/validators.

johanschvn:
if definied, set JWT secret to value of env variable JWTSECRET
closes #925 

Signed-off-by: Johan Schon <johanschvn@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ possible by the use of the `UserService` service provided by
 4. `return {token}` - send the JWT.
 
 You can see the details in
-[`packages/shopping/src/controllers/user-management.controller.ts`](https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/controllers/user.controller.ts).
+[`packages/shopping/src/controllers/user-management.controller.ts`](https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/controllers/user-management.controller.ts).
 
 ### Authorization
 
@@ -217,6 +217,17 @@ following rules:
 
 For more details about authorization in LoopBack 4, refer to
 https://loopback.io/doc/en/lb4/Loopback-component-authorization.html.
+
+### JWT secret
+
+By default, the JWTs will be signed using HS256 with a 64 character long string
+of random hex digits as secret. To use your own secret, set environment variable
+JWT_SECRET to the value of your own secret. You will want to use your own secret
+if running multiple instances of the application or want to generate or validate
+the JWTs in a different application.
+
+You can see the details in
+[`packages/shopping/src/application.ts`](./packages/shopping/src/application.ts).
 
 ### Tutorial
 
--- a/packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
+++ b/packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
@@ -4,7 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {Server} from '@grpc/grpc-js';
-import {TokenServiceConstants} from '@loopback/authentication-jwt';
+import {TokenServiceBindings} from '@loopback/authentication-jwt';
 import {securityId} from '@loopback/security';
 import {Client, expect} from '@loopback/testlab';
 import {
@@ -355,10 +355,8 @@ describe('UserController', () => {
    */
   async function givenAnExpiredToken() {
     const newUser = await createAUser();
-    const tokenService: JWTService = new JWTService(
-      TokenServiceConstants.TOKEN_SECRET_VALUE,
-      '-1',
-    );
+    const jwtSecret = app.getSync<string>(TokenServiceBindings.TOKEN_SECRET);
+    const tokenService: JWTService = new JWTService(jwtSecret, '-1');
     const userProfile = {
       [securityId]: newUser.id,
       name: `${newUser.firstName} ${newUser.lastName}`,
--- a/packages/shopping/src/application.ts
+++ b/packages/shopping/src/application.ts
@@ -22,6 +22,7 @@ import {
   RestExplorerComponent,
 } from '@loopback/rest-explorer';
 import {ServiceMixin} from '@loopback/service-proxy';
+import crypto from 'crypto';
 import fs from 'fs';
 import path from 'path';
 import {PasswordHasherBindings, UserServiceBindings} from './keys';
@@ -101,6 +102,12 @@ export class ShoppingApplication extends BootMixin(
 
     this.bind(UserServiceBindings.USER_SERVICE).toClass(UserManagementService);
     this.add(createBindingFromClass(SecuritySpecEnhancer));
+
+    // Use JWT secret from JWT_SECRET environment variable if set
+    // otherwise create a random string of 64 hex digits
+    const secret =
+      process.env.JWT_SECRET ?? crypto.randomBytes(32).toString('hex');
+    this.bind(TokenServiceBindings.TOKEN_SECRET).to(secret);
   }
 
   // Unfortunately, TypeScript does not allow overriding methods inherited


packages/shopping/src/application.ts
```diff
@@ -101,6 +101,16 @@ export class ShoppingApplication extends BootMixin(
 
     this.bind(UserServiceBindings.USER_SERVICE).toClass(UserManagementService);
     this.add(createBindingFromClass(SecuritySpecEnhancer));
+
+    // Use env variable JWTSECRET as secret key for JWT
+    const token = process.env.JWTSECRET;
+    if (!token) {
```
mrmodise, line 107: Thanks for the PR. I don't think its necessary to log an error rather just check that the `JWT_SECRET` exists and if it does use so that for those who dont want to set the environment variable are not affected by this change

mrmodise (changes requested):

packages/shopping/src/application.ts
johanschvn: The error message is there because I think using a publicly available key for signing JWTs is a very bad idea.
In production code I would have terminated the process if JWTSECRET was not set.

mrmodise: Interesting! maybe throwing an error would be more ideal based on what you mentioned above. Am just thinking if the variable is not set then it defaults back to the current behaviour where we use the default token?

johanschvn: Yes, if JWTSECRET is not set, you will get an error message, but other than that, it will work as before.
Changing to aborting program if JWTSECRET is not set will most probably break existing tests,
so I would not want to do that now.

nflaig: @johanschvn what about generating a secret via `crypto.randomBytes` if the env variable is not set?

If you run a single instance of the application this would be the most secure approach, however if multiple replicas of the app are used you will have to set the secret an env variable because all instances need the same secret

mrmodise: > If you run a single instance of the application this would be the most secure approach, however if multiple replicas of the app are used you will have to set the secret an env variable because all instances need the same secret

Thanks @nflaig. @johanschvn I think to keep things simple given the nature of the project you can generate a random string and leave a comment based on the above

Once these changes are done it will be ready to land on my side, again, thanks for the contributions

README.md
```diff
@@ -157,7 +157,18 @@ possible by the use of the `UserService` service provided by
 4. `return {token}` - send the JWT.
 
 You can see the details in
-[`packages/shopping/src/controllers/user-management.controller.ts`](https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/controllers/user.controller.ts).
+[`packages/shopping/src/controllers/user-management.controller.ts`](https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/controllers/user-management.controller.ts).
+
+### JWT secret key
+
+By default, the JWTs will be signed using HS256 with a 256 bit long random
+secret. To use your own secret, set environment variable JWTSECRET to the value
```
nflaig, line 165: @johanschvn should probably call the env variable `JWT_SECRET` which makes it easier to read and also follows the general naming convention

packages/shopping/src/application.ts
```diff
@@ -101,6 +102,11 @@ export class ShoppingApplication extends BootMixin(
 
     this.bind(UserServiceBindings.USER_SERVICE).toClass(UserManagementService);
     this.add(createBindingFromClass(SecuritySpecEnhancer));
+
+    // Use JWT secret from JWTSECRET environment variable if set
+    // otherwise create a random 256 bit long secret key
+    const secret = process.env.JWTSECRET || crypto.randomBytes(32).toString();
```
nflaig, line 108: @johanschvn should probably specify the encoding of the string either `toString("hex")` or `toString("base64")`

README.md
johanschvn: Agree, as currently used env variable RECOMMENDER_PROTOCOL follow that standard.

packages/shopping/src/application.ts
johanschvn: It is binary data - should be toString('binary'); I have changed. Using 'binary' instead of 'latin1' as I think it is easier to understand.

johanschvn: Actually, now I am not sure anymore what kind of string the library used (npm package jsonwebtoken) expect. I have only used keys with printable ASCII characters.

johanschvn: My best guess now is UTF8; maybe we should use 64 hex digits as the secret. Also I think I should remove "key" and only talk about secret, so not to confuse with public/private key pairs. What do you think?

mrmodise:
@johanschvn just 1 small thing on my end then we can land this from my end:

1. Squash commits
2. Move the JWT secret in README to after Authorization

Again, thanks for the contributions. Quick note on my end, we will merge once few more code owners approve the merge requests

packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
```diff
@@ -39,9 +38,11 @@ describe('UserController', () => {
   };
 
   const userPassword = 'p4ssw0rd';
+  const jwtSecret = 'a-secret-that-is-at-least-32-bytes-long';
   let expiredToken: string;
 
   before('setupApplication', async () => {
+    process.env.JWT_SECRET = jwtSecret;
```
mrmodise, line 45: Why is this necessary? ideally, nothing should break on the tests side based on the change you did since it was "backward" compatible

mrmodise (changes requested):

packages/shopping/src/application.ts
nflaig: The strength of the secret only depends on the size in bits but the encoding could potentially be necessary to avoid issues when passing the secret to another library. I usually use `base64` or `hex` encoding to prevent avoid those issues.

> Also I think I should remove "key" and only talk about secret, so not to confuse with public/private key pairs. What do you think?

just calling it secret sounds good to me

packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
johanschvn: It was backward compatible on my initial proposal, not after changing to have a random secret to be generated as default.
In the tests there is an expired key generated. It has to be generated with the same key as used by the application.
If random - it is unknown. Need to change tests to have a fixed known key.
Also, it is no longer fully backwards compatible; if having multiple instances and not setting JWT_SECRET, each instance will use a different secret; a key generated with one instance using random secret will not validate on other instances as they will have a different key.

packages/shopping/src/application.ts
johanschvn: Both changes are done: dropping "key" and encoding random key as "hex".

johanschvn:
@mrmodise in README.md, do you want me to move "JWT secret" to be between "Authorization" and "Tutorial" instead of between "Login" and "Authorization"?

mrmodise:
> @mrmodise in README.md, do you want me to move "JWT secret" to be between "Authorization" and "Tutorial" instead of between "Login" and "Authorization"?

Correct

packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
mrmodise: Okay will have a look once more. This approach feels hacky! Thanks for highlighting this

nflaig: The PR looks good now in my opinion. well done @johanschvn

@mrmodise I don't think the implementation is hacky, just addressing the points mentioned:

> If random - it is unknown. Need to change tests to have a fixed known secret.

That's a common approach to use a static/hardcoded secret during testing.

> Also, it is no longer fully backwards compatible; if having multiple instances and not setting JWT_SECRET, each instance will use a different secret; a secret generated with one instance using random secret will not validate on other instances as they will have a different secrets.

That is probably not a problem, I rather see it as a problem if someone uses the hardcoded secret of the module. And it makes sense to advocate good security practices in the example app as developers will use it as a starting point to build their own application which might be used in production in the end.

nflaig (approved):

packages/shopping/src/__tests__/acceptance/user-management.controller.acceptance.ts
mrmodise: @hacksparrow and @jannyHou your thoughts? I just think this would be more ideal solved at the security service model than to have multiple, probably, different implementations to achieve essentially the same thing.

I understand the common pattern of generating hard coded values on tests but wouldn't it rather paint a better picture from a testing perspective if we used a sample generated string from the crypo module?

johanschvn: > I just feel passing the hard coded string to an environment variable is unnecessary since ideally this should be managed outside of the application hence my thoughts on this being hacky. I mean this approach defeats the intended cause of good security practices

OK, I will try to change the test to instead read the randomly generated JWT secret from app context and use that for generating the expired token.

mrmodise: Cool

johanschvn: @mrmodise I have changed the test script now so it does not change/set environment JWT_SECRET; I still have to change the generation of the expired token; both the app and test code generating an expired token must use the same secret. I hope @nflaig likes this change too!?

nflaig (approved):

dougal83:
Personally I would use NODE_ENV environment variable to verbosely decide the source of subsequent values.  

This would:
- avoid further evaluation of the existence of environment value(s).  Such as `process.env.JWTSECRET || crypto.randomBytes(32).toString();`
- give flexibility to use numerous sets of values (say per environment)

Also, use of a set string is fine as generating a random string adds little value in tests.  Same goes for development environment if you ensure that the set string cannot be used in other environments such as production.

At the end of the day, sometimes the simplest solution is the best solution.

mrmodise:
> Personally I would use NODE_ENV environment variable to verbosely decide the source of subsequent values.
> 
> This would:
> 
> * avoid further evaluation of the existence of environment value(s).  Such as `process.env.JWTSECRET || crypto.randomBytes(32).toString();`

Agree, we could change the lines above to 

```ts
if (process.env.NODE_ENV === 'production') {
    this.bind(TokenServiceBindings.TOKEN_SECRET).to(crypto.randomBytes(32).toString('hex'));
}
```
> * give flexibility to use numerous sets of values (say per environment)
> 
> Also, use of a set string is fine as generating a random string adds little value in tests. Same goes for development environment if you ensure that the set string cannot be used in other environments such as production.

```ts
const tokenService: JWTService = new JWTService(
      TokenServiceConstants.TOKEN_SECRET_VALUE,
      '-1',
    );
```

> 
> At the end of the day, sometimes the simplest solution is the best solution.

@johanschvn you're the final voice on this 😄 . I think we presented a number of solutions by now

johanschvn:
@dougal83 I do not understand how evaluating environment variable NODE_ENV would simplify my initial problem of being able to select what secret to use for JWT signing. I need to be able to set the secret regardless of what environment the application is running in. How would you use NODE_ENV to select what value to use for the secret?

mrmodise:
> @dougal83 I do not understand how evaluating environment variable NODE_ENV would simplify my initial problem of being able to select what secret to use for JWT signing. I need to be able to set the secret regardless of what environment the application is running in. How would you use NODE_ENV to select what value to use for the secret?

See my comment above. by the way great work you're doing here. really appreciate it

johanschvn:
@mrmodise the code in your comment does not allow the secret to be set based on the value of env JWT_SECRET.
The value of TokenServiceConstants.TOKEN_SECRET_VALUE could not be used in my use case (a demo where the JWT in loopback4-example-shopping is signed by a different application). It is too short and is not allowed by the other program. See https://auth0.com/blog/brute-forcing-hs256-is-possible-the-importance-of-using-strong-keys-to-sign-jwts/ if you want to understand why you want to prevent the usage of short keys (like the other program I use in the demo do).

johanschvn:
Using a random key was not part of my initial issue or PR. It was implemented as an alternative to my initial proposal to log an error to warn about using an unsafe secret for JWT signing (not unsafe primarily because it was too short, but because it was publicly available).

mrmodise:
> @mrmodise the code in your comment does not allow the secret to be set based on the value of env JWT_SECRET.
> The value of TokenServiceConstants.TOKEN_SECRET_VALUE could not be used in my use case (a demo where the JWT in loopback4-example-shopping is signed by a different application). It is too short and is not allowed by the other program. See https://auth0.com/blog/brute-forcing-hs256-is-possible-the-importance-of-using-strong-keys-to-sign-jwts/ if you want to understand why you want to prevent the usage of short keys (like the other program I use in the demo do).

Thats just a prototype based on the comment. Feel free to do the actual implementation

johanschvn:
@mrmodise as for now I will not write any code using NODE_ENV as input for selecting JWT secret. @dougal83 or you have not yet convinced me it will have any general value or simplify the implementation for solving issue #925 . I currently do not see how I would be able to implement anything useful based on your prototype.

raymondfeng (approved):

dougal83:
> @mrmodise as for now I will not write any code using NODE_ENV as input for selecting JWT secret. @dougal83 or you have not yet convinced me it will have any general value or simplify the implementation for solving issue #925 . I currently do not see how I would be able to implement anything useful based on your prototype.

@johanschvn  Not a problem.  The issue being highlighted was addressing potential future issues if more variables are added.  Happy coding.

mrmodise (approved):

mrmodise:
yaay! finally landed @johanschvn. Lets land more

nflaig:
just as a further note, something like 

```ts
if (process.env.NODE_ENV === 'production') {
    this.bind(TokenServiceBindings.TOKEN_SECRET).to(crypto.randomBytes(32).toString('hex'));
}
```

is not practical because if you run multiple instances of the application all of them need the same secret else those instances would reject tokens created by another instance

PR #930, opened 2020-11-06
refactor: allow using any host and port on the frontend

### LINKED ISSUE #923: Shopping App Website: use other hosts than localhost
The Shopping App Website does not work (e.g. no products are displayed) if you try to run it on other hosts than localhost. In packages/shopping/public/config.js  the base URL of the API is coded as `const apiUrl = 'http://localhost:3000';`. Changing it to `const apiUrl = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '');` will make it work even if run on an other host or port.

johanschvn:
close #923

Signed-off-by: Johan Schon <johanschvn@gmail.com>

### DIFF
--- a/packages/shopping/public/config.js
+++ b/packages/shopping/public/config.js
@@ -5,7 +5,7 @@
 
 'use strict';
 
-const apiUrl = 'http://localhost:3000';
+const apiUrl = window.location.origin;
 const homePage = apiUrl + '/shoppy.html';
 const ordersPage = apiUrl + '/orders.html';
 const profilePage = apiUrl + '/profile.html';


packages/shopping/public/config.js
```diff
@@ -5,7 +5,11 @@
 
 'use strict';
 
-const apiUrl = 'http://localhost:3000';
+const apiUrl =
```
mrmodise, line 8: @dougal83 @hacksparrow @jannyHou what do you think?

dougal83: Oh I totally missed the point previously!  IMO just hard code in the case of the example but that looks like it will serve the purpose to make it dynamic.  Bear in mind the example is a bear bones demonstration, I don't think it was intended for production.

@hacksparrow is your man to give definitive answer. 👍

nflaig: @johanschvn I would recommend using `window.location.origin` instead

mrmodise: @nflaig +1 on window.location.origin. I think apiUrl is more descriptive since this is on the frontend side.

mrmodise:
@johanschvn you can do the change `window.location.origin` change then squash commits. I think it will be ready to land on my side. See https://loopback.io/doc/en/lb4/submitting_a_pr.html

Also notice I updated the title of the PR

Thanks, really appreciate the contributions.

mrmodise (approved):

mrmodise:
@johanschvn do another commit to trigger pipeline. Issue not related to your change.e.g `git commit --allow-empty -m "chore: re-trigger pipelines"`

raymondfeng (approved):

raymondfeng:
> @johanschvn do another commit to trigger pipeline. Issue not related to your change.e.g git commit --allow-empty -m "chore: re-trigger pipelines"

Let's squash the commits into one. We are ready to land it afterward.

mrmodise:
LGTM thanks @johanschvn

PR #934, opened 2020-11-09
feat: include OpenShift container build and deploy steps

johanschvn:
How to build and deploy loopback4-example-shopping on OpenShift Container Platform

### DIFF
--- a/openshift/Dockerfile.recommender
+++ b/openshift/Dockerfile.recommender
@@ -1,0 +1,12 @@
+FROM ubi8/nodejs-12
+
+USER 0
+RUN mkdir -p /home/node/app
+COPY base/packages/recommender /home/node/app
+RUN chown -R 1001:0 /home/node/app
+
+USER 1001
+WORKDIR /home/node/app
+ENV HOST=0.0.0.0 PORT=3001
+EXPOSE ${PORT} 50051
+CMD [ "node", "." ]
--- a/openshift/Dockerfile.shopping
+++ b/openshift/Dockerfile.shopping
@@ -1,0 +1,12 @@
+FROM ubi8/nodejs-12
+
+USER 0
+RUN mkdir -p /home/node/app
+COPY base/packages/shopping /home/node/app
+RUN chown -R 1001:0 /home/node/app
+
+USER 1001
+WORKDIR /home/node/app
+ENV HOST=0.0.0.0 PORT=3000
+EXPOSE ${PORT}
+CMD [ "node", "." ]
--- a/openshift/README.md
+++ b/openshift/README.md
@@ -1,0 +1,130 @@
+# Build and deploy on OpenShift Container Platform
+
+The loopback4-example-shopping/openshift directory contains Dockerfiles for
+building container images suitable for running on OpenShift Container Platform
+and yaml files for creating build configurations for building the images. They
+have been tested on OCP 4.6.9.
+
+The Dockerfiles in the loopback4-example-shopping/openshift directory differs
+from the Docker files in the root dirctory by using nodejs images based on ubi
+(Red Hat Universal Base Image) as base image, and by not using a hardcoded user
+named node. The node applications in the resulting images will be able to run
+using any userid.
+
+## First build and deploy
+
+The instructions here are intended to be used for a quick and easy installation
+for test and demo purposes only - not as an example of best practices for
+production use.
+
+The databases are set up without password protection. They will use emptyDir
+volumes for persistance, meaning that data will be lost if the pods are deleted
+or moved to an other node. Skip the `oc new-app mongo` and `oc new-app redis` if
+you want to set up databases in a different way.
+
+The application is exposed using http rather than https. No resource limits are
+set, which may inpact performance and ability to deploy to a cluster or project
+with enforced resource quotas.
+
+After logging in to your cluster using the `oc login` command and changing to
+the loopback4-example-shopping/openshift directory, run the commands below to
+create a new project called shoppy and to install the application and databases
+in the new project.
+
+```
+oc new-project shoppy
+oc new-app mongo --name shopping-app-mongodb
+oc new-app redis --name shopping-app-redis-master
+oc create imagestream shopping-base
+oc create imagestream shopping
+oc create imagestream recommender
+oc apply -f base.yaml
+oc apply -f recommender.yaml
+oc apply -f shopping.yaml
+```
+
+Wait for tag latest to be created for recommender and shopping. Use
+`oc get imagestreams` to see TAGS for imagestreams. Then run these commands:
+
+```
+oc new-app recommender --name recommender-rest
+oc new-app shopping
+oc expose svc shopping
+```
+
+Run `oc get route shopping` to get the hostname assigned to the application.
+
+## Start a new build and deploy
+
+If you want to do a new build and deploy after the repo has been updated, run:
+
+```
+oc start-build shopping-base-image
+```
+
+## Use your forked repo as source
+
+If you fork the strongloop/loopback4-example-shopping repo and want to use your
+own repo as source instead, you should edit base.yaml and change uri and ref:
+
+1. Change uri `https://github.com/strongloop/loopback4-example-shopping.git` to
+   the uri of your git repository.
+2. Change ref `master` to the branch of your repo that you want to build from.
+3. Then run these commands to update the buildconfig and start a new build:
+
+```
+oc apply -f base.yaml
+oc start-build shopping-base-image
+```
+
+## Configure SMTP-server and JWT_SECRET
+
+You can use environment variables to configure the SMTP server to be used for
+sending password reset emails and to set the secret to be used for generating
+JWTs. Example on how to set the environment variables is below. See root
+[README.md](../README.md) for details about each variable.
+
+```
+oc set env dc/shopping SMTP_PORT=587 \
+SMTP_SERVER=smtp.example.com \
+APPLICATION_URL=http://shopping-shoppy.mystack.example.com/ \
+SMTP_USERNAME=john@example.com \
+SMTP_PASSWORD=secret \
+PASSWORD_RESET_EMAIL_LIMIT=2 \
+JWT_SECRET=my-secret-is-exactly-32-bytes-lo
+```
+
+## Setting resource limits and node options
+
+You can set resource limits using the `oc set resources command`, e.g:
+
+```
+oc set resources dc shopping --limits=cpu=2000m,memory=2Gi \
+  --requests=cpu=500m,memory=1Gi`
+```
+
+To set options for node, you can set the NODE_OPTIONS environment variable, e.g:
+
+```
+oc set env dc/shopping NODE_OPTIONS=--max-old-space-size=1024
+```
+
+## Changing database connection details
+
+The databases created using the instructions under "First build and deploy" are
+not suitable for production. Creating databases in a more advanced way is
+outside the scope of this README.
+
+If you want to use databases that are password protected or set up with
+different names, in different projects or outside the cluster, you will have to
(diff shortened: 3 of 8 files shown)


raymondfeng:
@johanschvn Thank you for the contribution!

mrmodise:
Thanks @johanschvn just 1 comment on my side. When dealing with containers and node i think you want to restrict the [container memory and CPU ](https://phoenixnap.com/kb/docker-memory-and-cpu-limit)as well as [max-old-space-size](https://stackoverflow.com/questions/48387040/node-js-recommended-max-old-space-size)

The above will ensure better performance of your app and infrastructure

mrmodise:
@johanschvn this is really good. Looking forward to landing it

openshift/Dockerfile.recommender
```diff
@@ -0,0 +1,12 @@
...
+USER 0
+RUN mkdir -p /home/node/app
+COPY base/packages/recommender /home/node/app
+RUN chown -R 1001:0 /home/node/app
+
+USER 1001
+WORKDIR /home/node/app
+ENV HOST=0.0.0.0 PORT=3001
+EXPOSE ${PORT} 50051
+CMD [ "node", "." ]
```
mrmodise, line 12: Feel free to ignore. Let's add --max-old-space-size perhaps restrict to 1GB or so. 512 which is the default is very small IMO

openshift/Dockerfile.shopping
```diff
@@ -0,0 +1,12 @@
...
+USER 0
+RUN mkdir -p /home/node/app
+COPY base/packages/shopping /home/node/app
+RUN chown -R 1001:0 /home/node/app
+
+USER 1001
+WORKDIR /home/node/app
+ENV HOST=0.0.0.0 PORT=3000
+EXPOSE ${PORT}
+CMD [ "node", "." ]
```
mrmodise, line 12: Same comment on --max-old-space-size

openshift/README.md
```diff
@@ -0,0 +1,75 @@
+# Build and deploy on OpenShift Container Platform
+
+The loopback4-example-shopping/openshift contains Dockerfiles for building
+container images suitable for running on OpenShift Container Platform and yaml
+files for creating build configurations for building the images. They have been
+tested on OCP 4.6.9.
+
+## First build and deploy
+
+First run these commands in loopback4-example-shopping/openshift to create a new
```
mrmodise, line 10: do they have to be run in order? if yes maybe mention the same

openshift/Dockerfile.recommender
johanschvn: Not done in the Dockerfile.* in the root directory.
I do not want to add something that is not OpenShift related in this PR.

openshift/README.md
johanschvn: I think most people will try to run them in the order they are written and that will work.
I think it will add very little value explaining what commands can be run in a different order than listed.

mrmodise: I think we missed each other the recommendation was to mention that the comments should be executed in order. Remember not everyone who tries this out will be experienced in OpenShift. Change the line to:

```
First run the following commands in order in loopback4-example-shopping/openshift to create a new
```

openshift/Dockerfile.recommender
mrmodise: If thats the case I recommend you reference the root yaml than duplicate. What do you think?

johanschvn: The Dockerfile.* in root did not work for me in OpenShift.

mrmodise: Do you mind elaborating on this point?

johanschvn: Build terminates with this error if using ./Dockerfile.monorepo instead of openshift/Dockerfile.monorepo in base.yaml:
`error determining run uid: user: unknown user error looking up user "node"`

openshift/Dockerfile.shopping
johanschvn: I do not think you have to hard code --max-old-space-size in the image. If you want to specify --max-old-space-size, I think you can do it using the NODE_OPTIONS environment variable.

openshift/Dockerfile.recommender
mrmodise: Okay the users were different that means you probably running as root. Its discouraged but I know these Cloud platforms usually limit us

openshift/Dockerfile.shopping
mrmodise: Thanks, its actually possible. You will change your CMD command to the below:

RUN node --max-old-space-size=79000 .

You can even do more in the case of this project where we are using `dotenv` by passing the dotenv bootstrapping command like below

RUN node -r dotenv/config --max-old-space-size=1024 .

With the above being said, I think you need to update the last line to accommodate `dotenv`

mrmodise (changes requested):
For executing the CMD command on the Dockerfile.shopping you need accomodate the dotenv configuration by updating the CMD command to: 

CMD node -r dotenv/config --max-old-space-size=1024 .

openshift/Dockerfile.recommender
johanschvn: OpenShift by default runs containers with random userid for security reasons, not root.
Why do you think it is run as root?

openshift/Dockerfile.shopping
johanschvn: Last paragraph in my new README.md shows how to easily set environment variables in OpenShift.
The application appears to run fine without setting max-old-space-size. Why do you want to set it?
If you have some specific use case where it needs to be set you can do by setting environment variable NODE_OPTIONS to --max-old-space-size=79000 or whatever. I do not understand why you want it set when running in OpenShift, when it is not done otherwise!?

johanschvn:
@mrmodise, I do not agree that Dockerfile.shopping needs to be changed.
Are you running in OpenShift and experience errors when not using option --max-old-space-size=1024 ?

openshift/Dockerfile.recommender
johanschvn: Here is a blog post about userid in OpenShift:
https://cookbook.openshift.org/users-and-role-based-access-control/why-do-my-applications-run-as-a-random-user-id.html

openshift/Dockerfile.shopping
mrmodise: Thanks this is for supporting bootstrapping dotenv. You can see more information on this: https://www.npmjs.com/package/dotenv

mrmodise:
> @mrmodise, I do not agree that Dockerfile.shopping needs to be changed.
> Are you running in OpenShift and experience errors when not using option --max-old-space-size=1024 ?

Thanks however not experiencing errors does not mean we should not encourage best practices, as such, I recommend the change be applied before we land the PR.

johanschvn:
"-r dotenv/config" is not needed, environment variables are set by the `oc set env` command if you use my instructions.
If you want to start node with --max-old-space-size=1024, I think you can do so with `oc set env dc/shopping NODE_OPTIONS=--max-old-space-size=1024`, but I have not tested; and I am not sure how to test it.
Documentation for NODE_OPTIONS are here:
https://nodejs.org/docs/latest-v12.x/api/cli.html#cli_node_options_options

openshift/recommender.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  output:
+    to:
+      kind: ImageStreamTag
+      name: recommender:latest
+  source:
+    images:
+    - from:
+        kind: ImageStreamTag
+        name: shopping-base:latest
```
mrmodise, line 14: Please also apply memory and cpu usage limits at container level to avoid unnecessary wastage. See for reference: https://docs.openshift.com/container-platform/3.11/admin_guide/limits.html#overview

mrmodise (changes requested):

mrmodise:
> "-r dotenv/config" is not needed, environment variables are set by the `oc set env` command if you use my instructions.
> If you want to start node with --max-old-space-size=1024, I think you can do so with `oc set env dc/shopping NODE_OPTIONS=--max-old-space-size=1024`, but I have not tested; and I am not sure how to test it.
> Documentation for NODE_OPTIONS are here:
> https://nodejs.org/docs/latest-v12.x/api/cli.html#cli_node_options_options

Its configured in the project thus one can always create a `.env` file to pass the environment variables

openshift/README.md
```diff
@@ -0,0 +1,75 @@
...
+project called shoppy and to install the application and databases in the new
+project.
+
+```
+oc new-project shoppy
+oc new-app mongo --name shopping-app-mongodb
```
mrmodise, line 16: is this password protected?

mrmodise: Similarly, for Redis how does the password configuration work?

mrmodise:
@johanschvn alternative you can edit the `README.md` and highlight that this setup should not be used in production as-is without further security and performance optimisation configurations.

With that comment am happy to ignore the above and land the PR

johanschvn:
> @johanschvn alternative you can edit the `README.md` and highlight that this setup should not be used in production as-is without further security and performance optimisation considerations.
> 
> With that comment am happy to ignore the above and land the PR

@mrmodise thank you for your feedback
I agree that this is not an instruction for how to deploy to production.
I will add something about that this is a quick and easy way to get the example deployed into OpenShift for demo or test purpose and not production. I have put the PR in draft mode until ready for new review.
I am considering also adding more info:
* saying something about persistence - with this setup the databases are not persisted - if pod is moved to an other node, data will be lost
* maybe add instruction on how to expose the application as https instead of http  
* maybe add instructions on how to set resource limits
* maybe explaining why there are specific Dockerfile.* for building images for OpenShift

openshift/recommender.yaml
johanschvn: This should be in DeploymentConfig not in BuildConfig. I am considering adding info on how to do that.

openshift/README.md
johanschvn: There are no password protections for the databases.
The datasources expects that the databases are not password protected.
This is not something that is OpenShift specific, this is the current state of the application.
To start using passwords you have to change mango.datasource.ts and redis.datasource.ts.

The "--name shopping-app-mongodb" is for using the same solution as for Kubernetes example.

openshift/recommender.yaml
johanschvn: Added info about setting resource constraints to README.md in the current PR (force pushed with history rewritten).

mrmodise (approved):

mrmodise:
LGTM @johanschvn thanks for the PR as always. We can wait for one more approval then merge. 

PTAL @hacksparrow @jannyHou

mrmodise:
Closes #935

openshift/README.md
```diff
@@ -0,0 +1,130 @@
+# Build and deploy on OpenShift Container Platform
```
hacksparrow, line 1: A link to this doc from https://github.com/strongloop/loopback4-example-shopping/blob/master/README.md would be helpful. Please create an appropriately named section (inspiration https://github.com/strongloop/loopback4-example-shopping/blob/master/README.md#deploy-to-cloud-as-microservices).

@johanschvn @mrmodise

johanschvn: @hacksparrow @mrmodise what do you think about adding this after the Deploy to Cloud as Microservices section: 
```
## Build and deploy on Red Hat OpenShift

You can find [instructions](openshift/README.md), Dockerfiles and resource
definition files for building and deploying on Red Hat OpenShift Container
Platform in the openshift directory.
```

hacksparrow: LGTM

hacksparrow:
Thank you @johanschvn!

PR #973, opened 2020-12-17
feat: add password reset functionality and update forgot password

mrmodise:
This PR illustrates how password reset functionality can be enabled in a LoopBack 4 application using reset keys. A shopper who is locked out of their account will request a reset password link to be sent to their email where they will initiate the password reset process.


packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,41 @@
+import {bind, BindingScope} from '@loopback/core';
+const nodeMailer = require('nodemailer');
```
dougal83, line 2: Perhaps use `import nodemailer from 'nodemailer';`

packages/shopping/src/services/user-management.service.ts
```diff
@@ -83,4 +86,29 @@ export class UserManagementService implements UserService<User, Credentials> {
...
+  async requestPasswordReset(email: string): Promise<NodeMailer> {
+    const noAccountFoundError =
+      'No account associated with the provided email address.';
+    const foundUser = await this.userRepository.findOne({
+      where: {email},
+    });
+    if (!foundUser) {
+      throw new HttpErrors.NotFound(noAccountFoundError);
+    }
+    foundUser.resetKey = _.times(20, () => _.random(35).toString(36)).join('');
```
dougal83, line 99: Maybe use the following:
`import { v4 as uuidv4 } from 'uuid';`
`uuidv4();`

Continuing my anti lodash crusade. :)

dougal83 (commented):
Good start! 👍 👍

packages/shopping/src/services/email.service.ts
mrmodise: Types for nodemailer are not mature yet. Can be updated in future

packages/shopping/src/services/user-management.service.ts
mrmodise: I wouldn't install an additional package for such a logic. We already have lodash being used in the project and it works well 😄

dougal83: I think you'll find 'uuid' in the package-lock, correct me if I'm wrong.

packages/shopping/src/services/email.service.ts
dougal83: Even `https://www.npmjs.com/package/@types/nodemailer` not up to scratch?

mrmodise: You can propose usage and will take it from there. At the moment this is lower priority given the core functionality is incomplete

packages/shopping/src/services/user-management.service.ts
mrmodise: Sorry haven't checked but you can always verify and confirm the same

packages/shopping/src/services/email.service.ts
mrmodise: @dougal83 see https://github.com/nodemailer/nodemailer/issues/1094. Literally changing to import breaks the whole package setup hence my argument above

packages/shopping/src/services/user-management.service.ts
mrmodise: I checked and uuid is not installed and again I honestly don't see the need to install a new package for this. If perhaps concern is anti-lodash then we can use default JS `Math.random().toString(36).slice(2)`

packages/shopping/src/services/email.service.ts
dougal83: Ah, great thank you. So @types need an update. I looked at my old project and see that nodemailer was in sync with types at the time. Might see if I can contribute update.

packages/shopping/src/services/user-management.service.ts
dougal83: The change would remove the repeated random, which may cause confusion. The real reason for suggestion was that uuid v4 generated value creates random element that is not linked to timestamp. Lodash comment was really about me always trying to get rid of the expensive package.

Maybe better solution than v4 though:
https://medium.com/javascript-in-plain-english/you-might-not-need-uuid-v4-for-generating-random-identifiers-89e8a28a7d77

Maybe just use random and add comment to use something else in production?

packages/shopping/src/services/email.service.ts
dougal83, line 2: Types are fine. The error message could be more helpful.  Please install the types and try following changes, should work.  `npm i -D @types/nodemailer`

```suggestion
import { createTransport } from 'nodemailer';
```

packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,41 @@
...
+import {EmailTemplate, NodeMailer, User} from '../models';
+
+@bind({scope: BindingScope.TRANSIENT})
+export class EmailService {
+  private static async setupTransporter() {
+    return nodeMailer.createTransport({
```
dougal83, line 8:
```suggestion
    return createTransport({
```

packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,41 @@
...
+      host: process.env.SMTP_SERVER,
+      port: process.env.SMTP_PORT,
```
dougal83, line 10: This was the problem with type signature.  Always be weary of the types that you get from env. vars. in TypeScript.

```suggestion
      port: Number(process.env.SMTP_PORT) || 0,
```

packages/shopping/src/services/email.service.ts
dougal83: Types do not require update.  Error message was not clear.

dougal83 (changes requested):
Great stuff.  Up to you for random key; adding comment to tell others to use more refined solution in production should be good enough.

packages/shopping/src/services/email.service.ts
mrmodise: for port I would do `+(process.env.SMTP_PORT)` to covert to number

packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,41 @@
+import {bind, BindingScope} from '@loopback/core';
+import {EmailTemplate, NodeMailer, User} from '../models';
+import {createTransport} from 'nodemailer';
+
+@bind({scope: BindingScope.TRANSIENT})
+export class EmailService {
+  private static async setupTransporter() {
+    return createTransport({
+      host: process.env.SMTP_SERVER,
+      port: +(process.env.SMTP_PORT ?? 0),
```
dougal83, line 10: Note sure, might give NaN in some circumstances.  `+process.env.SMTP_PORT || 0`, I think would better?  If it tests OK, though... 👍

mrmodise: NaN is unavoidable since entering `SMTP_PORT=gccg` would result in NaN either with Number() or otherwise. I reused existing logic for PORT since its familiar at this point. `+process.env.SMTP_PORT ` wont work since the value is possibly undefined

packages/shopping/src/services/user-management.service.ts
nflaig: If you dont want to add `uuid ` I would recommend using the native `crypto.randomBytes` function directly. `uuid` uses this function internally to generate its random values because it is the best way to generate secure random values ([CSPRNG](https://gist.github.com/joepie91/7105003c3b26e65efcea63f3db82dfba)) in nodejs. Just need to ensure that the value is base64url encoded

```ts
const token = randomBytes(32)
        .toString("base64")
        .replace(/=/g, "")
        .replace(/\//g, "_")
        .replace(/\+/g, "-");
```

packages/shopping/src/services/email.service.ts
nflaig: wouldnt it be better to have `587` as the default value?

nflaig:
Really good implementation 👍

packages/recommender/package-lock.json
```diff
@@ -1,5326 +1,8 @@
 {
 	"name": "loopback4-example-recommender",
 	"version": "1.1.1",
-	"lockfileVersion": 2,
```
nflaig, line 4: @mrmodise is it intended that the content of all the package-lock files is removed?

mrmodise: not intended this file had conflicts earlier

packages/shopping/src/services/email.service.ts
mrmodise: to be honest any value here should be fine. When i was testing for example I was using my own SMTP server running on port 25 and Gmail runs on port 587. In any case there will be a specific error for wrong port so 0/587 is still fine to demonstrate that port is a number number value

dougal83: Avoidable NaN: [This is the way](https://www.typescriptlang.org/play?#code/MYewdgzgLgBNBOBLMBzAcgQzTAvDA5AEYZgkb4DcAsAFCiSwLLoCuAtrgQKyW233QY8AKYRM2PAGoAFE1TiYAfkUwADAEpqdcBBAAbYQDo9IFNIAGMueixKVGmAC4YAEgDeIsVgC+5zfx1YTzR2TisoJHlQ5TV-bUh9IxMzS1kI5hCOGIdnd2D2Xzi+eMFgrCgAC0QIAEkIABUK4QB1DABPMOsFAB9utS0BRONTC0ku2161J1cPUXFK6rrGlvbCgcChOfYF2oam1o6pLtDJ1XWEg2GUsfSojlPpvK22HaX91b8KIA)

packages/shopping/src/services/user-management.service.ts
dougal83: > I checked and uuid is not installed

I'm not sure where you checked but 'uuid' will be installed as per the package lock file.  Double checked and it is indeed found in 'node_modules' on install.  Please reconsider using uuid v4... it is a good option for this use case.

packages/shopping/src/services/email.service.ts
mrmodise: That example is different to our use case `process.env.SMTP_PORT` is not always a string but rather undefined hence `+undefined || 0` would always result in linting errors. I would really prefer to reuse existing implementation on this 1

packages/recommender/package-lock.json
mrmodise: @nflaig it looks like version 1 is the correct version. It gets changed when I install. See warning from pipeline and locally on this:

`npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
`

packages/shopping/src/services/email.service.ts
dougal83: I was trying to accommodate your preference for the `+` operator, so the original suggestion was the most suitable? `Number(process.env.SMTP_PORT) || 0`  As a bonus, the clarity would be self documenting without the need to know about the `+` trick.

mrmodise: Personally, to me the suitable suggestion would be to reuse the existing pattern

mrmodise:
> Great stuff. Up to you for random key; adding comment to tell others to use more refined solution in production should be good enough.

I see the changes are still outstanding. I thought this has been resolved let me know what other changes are needed on the uuid usage

packages/shopping/src/services/email.service.ts
mrmodise: @dougal83 resolving this one. Please open an issue or a PR to close off on this discussion.

packages/shopping/src/services/validator.ts
```diff
@@ -20,3 +22,26 @@ export function validateCredentials(credentials: Credentials) {
...
+  if (keyAndPassword.password !== keyAndPassword.confirmPassword) {
+    throw new HttpErrors.UnprocessableEntity(
+      'password and confirmation password do not match',
+    );
+  }
+
+  if (
+    _.isEmpty(keyAndPassword.resetKey) ||
+    keyAndPassword.resetKey.length === 0 ||
+    keyAndPassword.resetKey.trim() === ''
```
dougal83, line 43: Let's keep code consistent with L28 and lose the unnecessary lodash.

```suggestion
    !keyAndPassword.resetKey ||
    keyAndPassword.resetKey.trim().length === 0
```

packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+@bind({scope: BindingScope.TRANSIENT})
+export class EmailService {
+  /**
+   * If using gmail see https://nodemailer.com/usage/using-gmail/
+   */
+  private static async setupTransporter() {
+    return createTransport({
+      host: process.env.SMTP_SERVER,
+      port: +process.env.SMTP_PORT!,
```
dougal83, line 18: @jannyHou PTAL

dougal83 (approved):
OK.  Optional changes suggested.

mrmodise:
> OK. Optional changes suggested. Wait for one more approval.

Ya this is obviously a big change so it goes without saying that at least 2 approvals are needed

mrmodise:
FYI team, am writing an accompanying medium.com article on this.

packages/shopping/public/reset-password-finish.html
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+    <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js" integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
+    <script type="text/javascript" src="/purify.js"></script>
+    <script type="text/javascript" src="/templates.js"></script>
+    <script type="text/javascript" src="/config.js"></script>
+    <script type="text/javascript" src="/api.js"></script>
+    <script type="text/javascript" src="/util.js"></script>
+    <script type="text/javascript" src="/ui.js"></script>
+    <script type="text/javascript" src="/reset-password-finish.js"></script>
```
achrinza, line 25: Some remarks:

1. Add sub-resource integrity for js query
2. In accordance with WHATWG Living HTML Spec, `type` is not mandatory and should be removed as it's redundant. Also, `text/javascript` is deprecated in favour of `application/javascript`

```suggestion
    <script src="https://code.jquery.com/jquery-3.4.1.min.js" integrity="sha384-vk5WoKIaW/vJyUAd9n/wmopsmNhiy+L2Z+SBxGYnUkunIxVxAv/UtMOhba/xskxh" crossorigin="anonymous"></script>
    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js" integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
    <script src="/purify.js"></script>
    <script src="/templates.js"></script>
    <script src="/config.js"></script>
    <script src="/api.js"></script>
    <script src="/util.js"></script>
    <script src="/ui.js"></script>
    <script src="/reset-password-finish.js"></script>
```

packages/shopping/public/reset-password-init.html
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+    <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js" integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
+    <script type="text/javascript" src="/purify.js"></script>
+    <script type="text/javascript" src="/templates.js"></script>
+    <script type="text/javascript" src="/config.js"></script>
+    <script type="text/javascript" src="/api.js"></script>
+    <script type="text/javascript" src="/util.js"></script>
+    <script type="text/javascript" src="/ui.js"></script>
+    <script type="text/javascript" src="/reset-password-init.js"></script>
```
achrinza, line 25: Ditto

packages/shopping/public/reset-password-finish.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
```
achrinza, line 3: Missing mandatory charset:

```suggestion
  <head>
    <meta charset="utf-8">
```

packages/shopping/public/reset-password-init.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
```
achrinza, line 3: Ditto

```suggestion
  <head>
    <meta charset="utf-8">
```

packages/shopping/src/models/envelope.model.ts
```diff
@@ -0,0 +1,29 @@
...
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class Envelope extends Entity {
+  @property({
+    type: 'string',
+  })
+  from: string;
+
+  @property.apply({
```
achrinza, line 15: Could you clarify what `@property.apply()` is used for? I can't seem to find it in the LB4 docs.

achrinza (changes requested):
Great progress on the PR! I've left my initial remarks 👇

packages/shopping/src/models/envelope.model.ts
mrmodise: Weird. I think it is IDE auto complete.

packages/shopping/public/reset-password-finish.html
mrmodise: This is applied above

mrmodise:
@jannyHou @hacksparrow would be nice to have final approval from one of you. Please have a look when you are free. Thanks

packages/shopping/public/profile.html
```diff
@@ -4,24 +4,32 @@
     <title>
     Shoppy: The #1 online shopping destination
     </title>
+    <meta charset="utf-8">
     <link rel="icon" type="image/png" href="/favicon.png" />
```
achrinza, line 8:
```suggestion
    <link rel="icon" type="image/png" href="/favicon.png">
```

packages/shopping/public/profile.html
```diff
@@ -4,24 +4,32 @@
     <title>
     Shoppy: The #1 online shopping destination
     </title>
+    <meta charset="utf-8">
```
achrinza, line 7: Some browsers expect charset to be within the first nth bytes - Let's swap their positions and keep the title in a single line. Let's also add the viewport tag so that it'll render 1x zoom for mobile devices

```suggestion
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Shoppy: The #1 online shopping destination</title>
```

packages/shopping/public/index.html
```diff
@@ -3,7 +3,6 @@
 
 <head>
   <title>LoopBack 4 Example: Online Shopping APIs</title>
-
   <meta charset="utf-8">
```
achrinza, line 6: Ditto

```suggestion
    <meta charset="utf-8">
    <title>Shoppy: The #1 online shopping destination</title>
    <meta name="viewport" content="width=device-width, initial-scale=1">
```

packages/shopping/src/controllers/user-management.controller.ts
```diff
@@ -297,4 +306,70 @@ export class UserManagementController {
...
+    },
+  })
+  async resetPasswordFinish(
+    @requestBody() keyAndPassword: KeyAndPassword,
+  ): Promise<string> {
+    validateKeyPassword(keyAndPassword);
+
+    const foundUser = await this.userRepository.findOne({
+      where: {resetKey: keyAndPassword.resetKey},
+    });
```
achrinza, line 350: We might want to consider putting a time limit (e.g. 24 h) for the reset key to avoid a scenario where a user may have a password reset link with an "unlimited lifetime".

packages/shopping/public/product.html
```diff
@@ -4,10 +4,14 @@
     <title>
     Shoppy: The #1 online shopping destination
     </title>
+    <meta charset="utf-8">
```
achrinza, line 7: Ditto

```suggestion
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Shoppy: The #1 online shopping destination</title>
```

packages/shopping/public/reset-password-finish.html
```diff
@@ -0,0 +1,35 @@
...
+    <title>
+    Shoppy: The #1 online shopping destination
+    </title>
+    <meta charset="utf-8">
```
achrinza, line 7: Ditto

```suggestion
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Shoppy: The #1 online shopping destination</title>
```

packages/shopping/public/reset-password-init.html
```diff
@@ -0,0 +1,35 @@
...
+    <title>
+    Shoppy: The #1 online shopping destination
+    </title>
+    <meta charset="utf-8">
```
achrinza, line 7: Ditto

```suggestion
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Shoppy: The #1 online shopping destination</title>
```

packages/shopping/public/shoppy.html
```diff
@@ -4,10 +4,14 @@
     <title>
     Shoppy: The #1 online shopping destination
     </title>
+    <meta charset="utf-8">
```
achrinza, line 7: Ditto

```suggestion
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Shoppy: The #1 online shopping destination</title>
```

packages/shopping/src/models/email-template.model.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class EmailTemplate extends Entity {
```
achrinza, line 9: Since there's no ID, we should extend from `Model`. See [docs](https://loopback.io/doc/en/lb4/Model.html#overview) ([archive permalink](https://github.com/strongloop/loopback-next/blob/1300169a8e76d54abdf23559fadb8afa685ba76b/docs/site/Model.md#overview))

```suggestion
export class EmailTemplate extends Model {
```

achrinza (changes requested):
Great work on the PR! I've got some additional comments 👇

README.md
```diff
@@ -229,6 +239,25 @@ the JWTs in a different application.
...
 [`packages/shopping/src/application.ts`](./packages/shopping/src/application.ts).
 
+### Reset Password
+
+This repository includes a forgot password and reset password functionality that
+illustrates how shoppers can reset their password in the case they forgot them.
+Shoppers can either reset their password while logged in or locked out of the
+application. For this functionality we use Nodemailer. Please see
+https://nodemailer.com/usage/using-gmail/ if you're planning to use Nodemailer
+with GMAIL. Additionally, to manage environment variables we use `dotenv`,
```
hacksparrow, line 249: GMAIL -> Gmail

README.md
```diff
@@ -229,6 +239,25 @@ the JWTs in a different application.
...
+therefore, you must create a `.env` file in the root of the project with the
+below contents:
+
+```dotenv
+SMTP_PORT=587
+SMTP_SERVER=smtp.gmail.com
+APPLICATION_URL=http://localhost:3000 <endpoint-to-the-page-with-reset-password-form>
```
hacksparrow, line 256: `APPLICATION_URL=http://localhost:3000/<endpoint-to-the-page-with-reset-password-form>`

packages/shopping/src/controllers/user-management.controller.ts
```diff
@@ -297,4 +306,70 @@ export class UserManagementController {
...
+      },
+    },
+  })
+  async resetPasswordFinish(
+    @requestBody() keyAndPassword: KeyAndPassword,
+  ): Promise<string> {
+    validateKeyPassword(keyAndPassword);
+
+    const foundUser = await this.userRepository.findOne({
+      where: {resetKey: keyAndPassword.resetKey},
```
hacksparrow, line 349: Since `resetKey` defaults to "empty string" for newly created users and those that just reset their password, by making a request to `/users/reset-password/finish?resetKey=`, the account of the first user with an empty `resetKey` can be take over.

Once that account is taken over, we can initiate a password reset for that user to set the `resetKey` value for this user. A request to `/users/reset-password/finish?resetKey=` then will then yield the next user, etc.

Solution: always do `model.resetKey = uuidv4()`.

packages/shopping/src/controllers/user-management.controller.ts
hacksparrow: Yes. That would involve adding another property `passwordResetRequestTime` or something and then checking the time that has passed since it was requested, when the reset URL is called.

packages/shopping/src/controllers/user-management.controller.ts
hacksparrow: Oh, as I scrolled down, I could see empty `resetKey` is being checked. Please disregard the earlier comment.

mrmodise: @hacksparrow but we enforce a check for non empty reset key to avoid such a scenario on the validation

packages/shopping/src/controllers/user-management.controller.ts
```diff
@@ -297,4 +306,70 @@ export class UserManagementController {
...
+    },
+  })
+  async resetPasswordInit(
+    @requestBody() resetPasswordInit: ResetPasswordInit,
+  ): Promise<string> {
+    if (!isemail.validate(resetPasswordInit.email)) {
+      throw new HttpErrors.UnprocessableEntity('Invalid email address');
+    }
+
+    const nodeMailer: NodeMailer = await this.userManagementService.requestPasswordReset(
```
hacksparrow, line 324: Maybe put a limit on how many times to allow the reset request in an hour/day? Else, one could flood someone's email by making 100s of reset requests.

mrmodise: I will add another field on db table resetCount or something and limit it to 2. Does that sound like a sound solution?

hacksparrow: Yes two per hour sounds good. Also you will have to show an appropriate message when this limit is reached.

mrmodise: @hacksparrow I think am missing something. You would prefer to limit the reset emails to 2 per hour in a day? or to 2 emails a day?

hacksparrow: Any would do, as long as we have a configurable setting.

hacksparrow (approved):
LGTM. Please address @achrinza's change requests too.

mrmodise:
> LGTM. Please address @achrinza's change requests too.

Already addressed and requested review

packages/shopping/src/controllers/user-management.controller.ts
mrmodise: Included

mrmodise: @achrinza please confirm if you are happy with these changes

packages/shopping/src/services/user-management.service.ts
```diff
@@ -83,4 +88,74 @@ export class UserManagementService implements UserService<User, Credentials> {
...
+  async updateResetRequestLimit(user: User): Promise<User> {
+    const resetTimestampDate = new Date(user.resetTimestamp);
+
+    const difference = await subtractDates(resetTimestampDate);
+
+    if (difference === 0) {
+      user.resetCount = user.resetCount + 1;
+
+      if (user.resetCount > +(process.env.PASSWORD_RESET_EMAIL_LIMIT ?? 2)) {
+        throw new HttpErrors.BadRequest(
```
achrinza, line 128:
```suggestion
        throw new HttpErrors.TooManyRequests(
```

packages/shopping/src/controllers/user-management.controller.ts
```diff
@@ -97,6 +105,7 @@ export class UserManagementController {
     validateCredentials(_.pick(newUserRequest, ['email', 'password']));
 
     try {
+      newUserRequest.resetKey = '';
```
achrinza, line 108: I'm wondering if we should the more semantically-correct database NULL instead of a blank string.

packages/shopping/src/services/email.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    const emailTemplate = new EmailTemplate({
+      to: user.email,
+      subject: '[Shoppy] Reset Password Request',
+      html: `
+      <div>
+          <p>Hello, ${user.firstName} ${user.lastName}</p>
+          <p style="color: red;">We received a request to reset the password for your account with email address: ${user.email}</p>
+          <p>To reset your password click on the link provided below</p>
+          <a href="${process.env.APPLICATION_URL}/reset-password-finish.html?resetKey=${user.resetKey}">Reset your password link</a>
+          <p>If you didn’t request to reset your password, please ignore this email or reset your password to protect your account.</p>
```
achrinza, line 37: Do we sanitize the user's first and last names? Otherwise, this may be an opportunity for Self-XSS.

Feel free to ignore this as it's an issue that's scoped for the entire application rather than just the reset password feature.

packages/shopping/src/services/user-management.service.ts
```diff
@@ -9,17 +9,22 @@ import {repository} from '@loopback/repository';
...
+import {v4 as uuidv4} from 'uuid';
+import {subtractDates} from '../utils';
 
 export class UserManagementService implements UserService<User, Credentials> {
   constructor(
     @repository(UserRepository)
     public userRepository: UserRepository,
     @inject(PasswordHasherBindings.PASSWORD_HASHER)
     public passwordHasher: PasswordHasher,
+    @inject('services.EmailService')
```
achrinza, line 26: Preferably, we shouldn't need to know the exact binding key when doing dependency injection (separation of responsibilities). So either we derive the binding key from the class _or_ create a binding key in `keys.ts` like the password hasher above this.
```suggestion
    @service(EmailService)
```

achrinza (approved):
Thanks for working through the changes! I have some non-blocking comments. Feel free to ignore them.

packages/shopping/src/services/email.service.ts
mrmodise: Yep! we will need to do another PR for this.

packages/shopping/src/services/user-management.service.ts
mrmodise: I think this needs another PR to fix since the pattern already exists. See `product.controller.ts` and `user-management.controller.ts`. The case for password hasher is a unique one since this is a custom service that doesn't use default binding attributes of the framework

packages/shopping/src/controllers/user-management.controller.ts
mrmodise: I think this is fine for a NoSQL database. If we were using SQL database it would make more sense to

PR #977, opened 2020-12-19
feat: migrate sequence to middleware-based

mrmodise:
The default Sequence has been migrated from Action-based sequence to Middleware-based. This PR upgrades the default project Sequence to middleware-based so that users can benefit from how this new approach works


packages/shopping/src/middlewares/shoppy.middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+    // mappings from error codes to HTTP status codes, so that application
+    // don't have to map codes themselves.
+    if (
+      error.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
+      error.code === USER_PROFILE_NOT_FOUND
+    ) {
+      Object.assign(error, {statusCode: 401 /* Unauthorized */});
+    }
+    reject(ctx, error);
+  }
```
achrinza, line 46: Any reason for this middleware? AFAIK, this is already handled by the Authentication middleware:

https://github.com/strongloop/loopback-next/blob/883e9b31b0e5281085a0c9f091f648e84fef795e/packages/authentication/src/providers/auth-action.provider.ts#L130-L141

achrinza (changes requested):
Just need a clarification on the Middleware. Other than that, everything else LGTM 👍

packages/shopping/src/middlewares/shoppy.middleware.ts
mrmodise: Good catch. To be honest am looking at @raymondfeng or anyone who has more exposure to middleware to weigh in. One of the reasons why I did this change I feel like for most users its still a grey area myself included

nflaig:
Just a general note, I am not sure if I like the middleware sequence compared to the previous sequence. It is no longer transparent to the user how requests are actually processed which was quite nice in the action based sequence. Of course having middleware based sequence allows to automatically add things like authentication which is quite nice.

What are your thoughts on this?

mrmodise:
> Just a general note, I am not sure if I like the middleware sequence compared to the previous sequence. It is no longer transparent to the user how requests are actually processed which was quite nice in the action based sequence. Of course having middleware based sequence allows to automatically add things like authentication which is quite nice.
> 
> What are your thoughts on this?

Same here it took me a while to migrate to it. It wasn't clear how the Sequence and the middleware work together hence I updated the docs: https://github.com/strongloop/loopback-next/pull/6824

packages/shopping/src/middlewares/shoppy.middleware.ts
raymondfeng: If you want to customize the error mapping, please check out an example at https://github.com/strongloop/loopback-next/blob/master/examples/validation-app/src/middleware/validation-error.middleware.ts.

mrmodise: Thanks @raymondfeng I will apply it on my projects and update PR accordingly

mrmodise: Didn't find a use case for the middleware provider on this repo. However, it did help me on my projects at work. I will remove the customisation for now until i find a way i can illustrate this better for our users

packages/shopping/src/middlewares/shoppy.middleware.ts
```diff
@@ -0,0 +1,16 @@
...
+  next: Next,
+) => {
+  const send = await ctx.get(RestBindings.SequenceActions.SEND);
+  const reject = await ctx.get(RestBindings.SequenceActions.REJECT);
+  try {
+    const result = await next();
+    send(ctx.response, result);
+  } catch (error) {
+    reject(ctx, error);
+  }
```
achrinza, line 15: This looks identical to the [SendResponseMiddlewareProvider](https://github.com/strongloop/loopback-next/blob/1fe4c274efb32210699c13be38d8b71365bee9c4/packages/rest/src/providers/send.provider.ts#L42-L55), which is included in the [default REST Middleware Chain](https://github.com/strongloop/loopback-next/blob/1fe4c274efb32210699c13be38d8b71365bee9c4/packages/rest/src/sequence.ts#L195-L198).

Since the default Middleware Sequence is used, there shouldn't be a need for a custom middleware.

achrinza (changes requested):

packages/shopping/src/middlewares/shoppy.middleware.ts
mrmodise: Closing PR since I won't have time to look into this at the moment

mrmodise: @achrinza the objective is to show the use of the middleware hence I was leaning on @raymondfeng providing a use case for such as well as for the provider. Most users don't understand how these fit together even from the official documentation it is not clear. Looking at how we missed this the first time I assume we are also not fully familiar with how this works

mrmodise: @raymondfeng please help with a use case, which I can then use to update the documentation on this topic. Am sure from an architecture perspective there was a good reason to switch from a default action-based sequence to this new approach. I would like to make it clear to our users

achrinza: I think it might add to the confusion as readers may assume that this is a necessary step, when in reality, it's not necessary as there's already a default middleware provider handling this. The concern is that developers may blindly copy this even though it's not the "correct" usage.

In this case, improving the documentation would be the more effective solution.

achrinza: I think it might add to the confusion as readers may assume that this is a necessary step, when in reality, it's not necessary as there's already a default middleware provider handling this. The concern is that developers may blindly copy this even though it's not the "correct" usage.

In this case, improving the documentation would be the more effective solution.

mrmodise: I think we missed each other. What am referring to on the above is I want to demonstrate how the Sequence, Middleware and possibly middleware-provider fit together not necessarily use the above changes

mrmodise: @achrinza I found how the error handling is done in the middleware-based sequence. Instead of recreating the middleware like I did earlier one rather creates a Middleware-provider that somewhat has a middleware in it. See the updated changes. I have to admit its a lot though unlike the previous action-based. I

Also, I figured out why we were missing each other. My fault, I kind of assumed we were on the same page. Context, we had error handling implemented in the action-based sequence, as such when I was migrating to middleware-base sequence I also wanted to demonstrate how does one handle errors. Trust me this can be a daunting task I personally went through it and based on Raymond’s link above I just discovered I went about it the wrong way. The correct way is in the updated changes
 
PTAL @raymondfeng

mrmodise:
@achrinza PTAL. I can confirm this to be correct having done similar changes for our apps in production, where we are using the error handler middleware for logging all Errors to Sentry

achrinza (approved):
Apologies for the delayed response; Thanks for adding the latest changes, everything LGTM 👍

mrmodise:
@hacksparrow PTAL

PR #978, opened 2020-12-19
feat: remove index.js and add optional rest configs

mrmodise:
This PR upgrades this repository to the new Lb4 setup structure of not having index.js in the setup. Furthermore,  it also illustrates how to setup further configs like CORS and Ciphers in a Lb4 app


achrinza (approved):
LGTM 👍

PR #989, opened 2020-12-28
refactor: delete user credentials on initial data migration

acasperw:


mrmodise:
@acasperw nice catch and thanks for the PR. Please see https://loopback.io/doc/en/lb4/submitting_a_pr.html#how-to-submit-a-pull-request-pr-for-loopback-4 on how to submit a PR to this repository to avoid pipeline failing.

acasperw:
@mrmodise , thanks, iv added the DCO (never used it before), i can see the sign off in my commits but it still fails?

mrmodise:
> @mrmodise , thanks, iv added the DCO (never used it before), i can see the sign off in my commits but it still fails?

I think you missed some steps from that link: https://loopback.io/doc/en/lb4/submitting_a_pr.html#9-final-rebase-and-squashing-of-commits

dougal83:
I'd probably go with combining all commits into one called `feat: delete user credentials on initial data migration` as you're adding a function to the code.

PR #1000, opened 2021-01-10
ci: add appveyor CI

mrmodise:
I cross-checked with loopback-next repo and Appveyor is still being used. Adding it back for this repo too


achrinza:
loopback-next still has AppVeyor as we weren't able set-up the GitHub Actions Windows runner yet due to tty issues with the CLI tests.

IMO it should be fine to get rid of AppVeyor from this repo since GH Actions is able to replace all the tests.

PR #1278, opened 2021-08-06
use proper svc name

xw-patrick:
service is called shopping (not shopping-app)
(at least when I deployed it locally on my machine ;-))


mrmodise:
Closing due to inactivity. Feel free to open when ready. Thanks

